PR #7, opened 2016-04-07
Assert correct header format during construction

qinfchen:
I spoke too soon. Original change was correct that it is case-insensitive
https://tools.ietf.org/html/rfc2617#section-1.2

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
@@ -16,6 +16,9 @@
 
 package com.palantir.tokens.auth;
 
+import com.google.common.base.Preconditions;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.immutables.value.Value;
 
 /**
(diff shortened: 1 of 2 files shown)


qinfchen:
@pnepywoda @ellisjoe

auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
```diff
@@ -16,25 +16,40 @@
...
 import org.immutables.value.Value;
 
 /**
  * Represents a HTTP authentication header. This class wraps a string in the form of "Bearer [token]".
  */
 @Value.Immutable
 @Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE, jdkOnly = true)
-// NOTE: no @JsonSerialize/@JsonDeserialize because auth headers are for use in @HeaderParam
-// see: https://jersey.java.net/apidocs/latest/jersey/javax/ws/rs/HeaderParam.html
+@JsonSerialize(as = ImmutableAuthHeader.class)
```
pnepywoda: -1
what happened here?

qinfchen: oops, fixed

markelliot:
is there still interest in moving forward this PR?

qinfchen:
@pnepywoda and @ellisjoe?

auth-tokens/src/test/java/com/palantir/tokens/auth/AuthHeaderTest.java
```diff
@@ -32,6 +32,31 @@ public void testSimple() {
...
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testMissingBearer() {
+        AuthHeader.valueOf("someToken");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testMultipleTokens() {
+        AuthHeader.valueOf("Bearer tokenOne tokenTwo");
```
markelliot, line 47: it seems like this fails due to being a bad token format, not because it doesn't match the regex

auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
```diff
@@ -27,14 +30,24 @@
 // see: https://jersey.java.net/apidocs/latest/jersey/javax/ws/rs/HeaderParam.html
 public abstract class AuthHeader {
 
+    // https://tools.ietf.org/html/rfc2617#section-1.2 for case insensitive auth-scheme
+    private static final String VALIDATION_PATTERN_STRING = "^bearer\\s+(.*)$";
```
markelliot, line 34: it may be faster to use http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#regionMatches%28boolean,%20int,%20java.lang.String,%20int,%20int%29 instead of regex. Not sure it matters.

qinfchen, line 34: +1

qinfchen:
I will close this PR for now. We can revisit this if need to.

PR #9, opened 2016-05-16
Make it possible to extract the user id from a bearer token

### LINKED ISSUE #8: Provide method for getting user id from header
Hey all,
We've found it useful to have code that extracts the 'sub' field from a JWT auth header.
Obviously we can't trust this, but it's useful for logging - being able to say that a certain user made a given request is useful without having to pull in all the complexity of an auth bundle.

Would a `getUserIdInsecure()` method be welcome as part of the AuthHeader or the BearerToken class?

j-baker:
Fixes #8.

Additionally, see internal bug #1165.

This code is kinda Palantir specific, but it doesn't restrict the use of this package (in an open source context).

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
@@ -18,7 +18,15 @@ package com.palantir.tokens.auth;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
+import com.google.common.io.BaseEncoding;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.UUID;
 import java.util.regex.Pattern;
 import org.immutables.value.Value;
 
@@ -31,6 +39,47 @@ public abstract class BearerToken {
 
     private static final String VALIDATION_PATTERN_STRING = "^[A-Za-z0-9\\-\\._~\\+/]+=*$";
     private static final Pattern VALIDATION_PATTERN = Pattern.compile(VALIDATION_PATTERN_STRING);
+
+    private static final String SUBJECT_FIELD = "sub";
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    /**
+     * If the token contains a 'sub' field, as a base64 encoded UUID, return it as a String.
+     * @return the 'sub' field of the JWT, if it exists
+     */
+    public final Optional<String> getUserIdInsecure() {
+        try {
+            String[] segments = getToken().split("\\.");
+
+            if (segments.length != 3) {
+                return Optional.absent();
+            }
+
+            String b64EncodedJson = segments[1];
+
+            String json = new String(BaseEncoding.base64().decode(b64EncodedJson), StandardCharsets.UTF_8);
+
+            JsonNode node = objectMapper.readTree(json);
+
+            if (!node.has(SUBJECT_FIELD)) {
+                return Optional.absent();
+            }
+
+            return Optional.of(uuidFromBytes(
+                    BaseEncoding.base64().decode(node.get(SUBJECT_FIELD).asText())).toString());
+        } catch (RuntimeException | IOException e) {
+            return Optional.absent();
+        }
+    }
+
+    private static UUID uuidFromBytes(byte[] bytes) {
+        ByteBuffer byteBuffer = ByteBuffer.wrap(bytes);
+
+        long high = byteBuffer.getLong();
+        long low = byteBuffer.getLong();
+
+        return new UUID(high, low);
+    }
 
     @Value.Parameter
     @JsonValue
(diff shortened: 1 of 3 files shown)


auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -32,6 +40,47 @@
...
     private static final Pattern VALIDATION_PATTERN = Pattern.compile(VALIDATION_PATTERN_STRING);
 
+    private static final String SUBJECT_FIELD = "sub";
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    /**
+     * If the token contains a 'sub' field, as a base64 encoded UUID, return it as a String.
+     * @return the 'sub' field of the JWT, if it exists
+     */
+    public final Optional<String> getUserIdInsecure() {
```
ellisjoe, line 50: a BearerToken isn't always a JWT is it? feels like this should go into a JWT util class

j-baker, line 50: The ones that are used by Palantir are, right? I don't think we need to compromise for non-Palantir use cases.

Beyond that, I think that this is fine - if it's not a JWT in the format generated by our internal auth solution, you just don't get a user id back.

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -32,6 +40,47 @@
...
+        try {
+            String[] segments = getToken().split("\\.");
+
+            if (segments.length != 3) {
+                return Optional.absent();
```
markelliot, line 55: Probably ought to log something here?

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -32,6 +40,47 @@
...
+            JsonNode node = objectMapper.readTree(json);
+
+            if (!node.has(SUBJECT_FIELD)) {
+                return Optional.absent();
+            }
+
+            return Optional.of(uuidFromBytes(
+                    BaseEncoding.base64().decode(node.get(SUBJECT_FIELD).asText())).toString());
+        } catch (RuntimeException | IOException e) {
+            return Optional.absent();
```
markelliot, line 71: And here

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -32,6 +40,47 @@
...
+            }
+
+            String b64EncodedJson = segments[1];
+
+            String json = new String(BaseEncoding.base64().decode(b64EncodedJson), StandardCharsets.UTF_8);
+
+            JsonNode node = objectMapper.readTree(json);
+
+            if (!node.has(SUBJECT_FIELD)) {
+                return Optional.absent();
```
markelliot, line 65: And here

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
markelliot, line 50: It seems like it'd generally be cleaner to have this just extract the "sub" field from a JWT, and for the consumer to decide if they want to render as a UUID.

j-baker, line 50: I think the disadvantage there is that every single consumer is going to write/use this transformation.

The question is whether this bearer token is for Palantir bearer tokens, or whether it's for generic JWTs in order to help third-party users. If we view it to be the former then we should make it return a uuid formatted string (or maybe even an actual Java UUID). If we view it to be the latter, then we should just return the 'sub' field, as you suggest.

My view is that we should do the former. This class has only been open sourced because @qinfchen wanted to use it in http-remoting, and has little value to anyone else (in particular, its value within Palantir is solely due to network effects (you don't need to convert your bearer tokens between services)). I don't think that we should compromise usage within Palantir for this.

By compromise, here I mean that I want to write

```
logger.warn("User {} tried to write to {} but had no permission", bearerToken.getUserIdInsecure(), dataset);
```

and not

```
logger.warn("User {} tried to write to {} but had no permission",
    UserIdFormatter.format(bearerToken.getUserIdInsecure()), dataset);
```

by which point we might as well just do

```
logger.warn("User {} tried to write to {} but had no permission",
    UserIdFormatter.format(bearerToken), dataset);
```

The reason why I think that the latter case is worse is because UserIdFormatter is less discoverable and has more boilerplate. Our users will have to exclusively use a hypothetical UserIdFormatter because the base-64 encoding is an artifact of our JWTs only, and is not discoverable in any of our UIs.

Does this make sense?

markelliot, line 50: I'm in sync that this library can and should likely be Palantir specific, but I don't think the functionality encoded here belongs in this method on this class.

We likely want to parse out other JWT content, such as the sessionId, and also the expiration time, and it seems reasonable that in keeping with this project's semantic and narrow definitions to add a method like `JsonWebToken asJwt()` to this class, then to stick parsing of `userId`, `sessionId` and `expiration` into `JsonWebToken`.

Your logging examples wouldn't be much more complex:

```
logger.warn("User {} ...",
    bearerToken.asJwt().getUserId())
```

Separately, we should likely consider caching some state here so that we're not re-parsing the whole thing if we want to render multiple fields:

```
logger.warn("User {} with Session {} ...",
    bearerToken.asJwt().getUserId(),
    bearerToken.asJwt().getSessionId())
```

As an alternative, if we give the JWT class a nice stringify/logger function, we can likely normalize how UserId and SessionId are written to the logs, which seems advantageous.

j-baker, line 50: Good idea, thanks

markelliot:
@j-baker were you planning on revisiting this or should we get someone else to take a stab?

markelliot:
Fixed in #10

PR #10, opened 2016-06-08
Add support for parsing unverified JWTs

markelliot:


markelliot:
Not sure why I'm not getting a build =/

markelliot:
@uschi2000 @pnepywoda @j-baker

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -49,4 +58,8 @@ public static BearerToken valueOf(String token) {
     public final String toString() {
         return getToken();
     }
+
+    public final JsonWebToken asJwt() {
```
quellkot: so is every bearertoken a JWT? if so, then why not bake this assumption into the BearerToken directly?

auth-tokens/src/main/java/com/palantir/tokens/auth/JsonWebToken.java
```diff
@@ -0,0 +1,67 @@
...
+public abstract class JsonWebToken {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper()
+            .registerModule(new GuavaModule());
+
+    /**
+     * Returns an unverified user id.
+     */
+    @Value.Parameter
+    public abstract String getUnsafeUserId();
```
quellkot: what's the use-case for this? IMO, seems pretty dangerous to expose such a method. sooner or later, it'll be abused.

auth-tokens/src/main/java/com/palantir/tokens/auth/JsonWebToken.java
```diff
@@ -0,0 +1,67 @@
...
+        JsonWebTokenPayload payload = extractPayload(segments[1]);
+
+        return ImmutableJsonWebToken.of(decodeUuidBytes(payload.getSub()));
+    }
+
+    private static JsonWebTokenPayload extractPayload(String payload) {
+        try {
+            return MAPPER.readValue(BaseEncoding.base64().decode(payload), JsonWebTokenPayload.class);
+        } catch (IOException e) {
+            throw new RuntimeException("Invalid JWT: cannot parse payload", e);
```
quellkot: IllegalArgumentException?

uschi2000:
sorry, signed in with wrong account. quellkot == uschi2000

auth-tokens/src/main/java/com/palantir/tokens/auth/JsonWebToken.java
markelliot: Primarily for logging -- we'd like services to log user id without necessarily having to cryptographically verify (as a best effort kind of thing). We could forcibly add a string "-untrusted" to make it more obvious or mark the method as deprecated to discourage use. 

Some more discussion on
https://github.com/palantir/auth-tokens/issues/8

quellkot: I see. I think I'd prefer to make this a static method like
`BearerTokens#extractUnverifiedUserId(BearerToken token)` just so it's one
step further away. Thoughts?

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
markelliot: It could be, asJwt will throw if it's not. There's some more discussion of this approach here: https://github.com/palantir/auth-tokens/pull/9

If you prefer the getter on BearerToken we can go with #9 instead, but I think this makes it a bit clearer what the setup is and gives us the opportunity to reveal other encoded/untrusted data (as well as provide some options for indicating trust or offering to verify things). 

I did add the optional fields in for the payload parser, we could also expose those but I couldn't decide on how to differentiate between session and API tokens reasonably.

auth-tokens/src/main/java/com/palantir/tokens/auth/JsonWebToken.java
markelliot: Happy to remove the asJwt method and switch to doing something like:

```
JsonWebToken.of(BearerToken).getUnverifiedUserId()
```

Can also rename JsonWebToken to UnverifiedJsonWebToken to drive the point home.

uschi2000: yep, I think that's much better. (it also removes the implicit assumption from BearerToken that's it's always a JWT)

markelliot: Ok, pushed a refactor.

We also need to add a servlet filter that pushes this into the slf4j MDC (http://logback.qos.ch/manual/mdc.html) -- so another option is to move this to that project, but I think it might be useful to key on `sid` or `jti` for cache keys.

auth-tokens/src/main/java/com/palantir/tokens/auth/JsonWebToken.java
markelliot: done

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -49,4 +49,5 @@ public static BearerToken valueOf(String token) {
     public final String toString() {
         return getToken();
     }
+
```
uschi2000: boing

auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
```diff
@@ -0,0 +1,80 @@
...
+ */
+@Value.Immutable
+@Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE, jdkOnly = true)
+public abstract class UnverifiedJsonWebToken {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper()
+            .registerModule(new GuavaModule());
+
+    /**
+     * Returns an unverified user id.
```
uschi2000: "Returns the unverified user id, i.e., the "sub" (subject) field of the JWT"

auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
```diff
@@ -0,0 +1,80 @@
...
+    private static JsonWebTokenPayload extractPayload(String payload) {
+        try {
+            return MAPPER.readValue(BaseEncoding.base64().decode(payload), JsonWebTokenPayload.class);
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Invalid JWT: cannot parse payload", e);
+        }
+    }
+
+    private static String decodeUuidBytes(byte[] bytes) {
+        checkArgument(bytes.length == 16, "Invalid JWT: cannot decode UUID, require 16 bytes, found %s", bytes.length);
```
uschi2000, line 84: is this a private convention?

auth-tokens/src/test/java/com/palantir/tokens/auth/JsonWebTokenTests.java
```diff
@@ -0,0 +1,73 @@
...
+package com.palantir.tokens.auth;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+import org.junit.Test;
+
+public final class JsonWebTokenTests {
+
+    private static final BearerToken SESSION_TOKEN = BearerToken.valueOf("eyJhbGciOiJFUzI1NiJ9."
```
uschi2000, line 26: this is a little silly, but I guess it's overkill to include a JWT library to generate these tokens from plaintext...

auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
markelliot, line 84: It's Palantir-centric, yes.

auth-tokens/src/test/java/com/palantir/tokens/auth/JsonWebTokenTests.java
markelliot, line 26: I can include the plain text as a comment if you'd like.

auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
uschi2000, line 84: in other words, should this fallback to the `char[]/string` representation of these bytes if it's not a UUID?

uschi2000:
looks ok modulo question around UUID

auth-tokens/src/test/java/com/palantir/tokens/auth/JsonWebTokenTests.java
uschi2000, line 26: not necessary

auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
markelliot, line 84: Might be a bit weird if someone has a 16 byte string for sub/there's no easy way to distinguish things that are valid/invalid UUIDs.

quellkot, line 84: true. let's leave it as is.

uschi2000:
I don't have merge permissions here, but I'm +1

auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
markelliot, line 84: added a comment

auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
markelliot: done

markelliot:
@uschi2000 you should have push rights now.

pnepywoda:
+1

uschi2000:
can someone with admin rights check the circlici integration, please?

uschi2000:
@pnepywoda @markelliot

PR #11, opened 2016-06-10
Add a Filter for augmenting logging context with user details

markelliot:


gradle/publish-libs.gradle
```diff
@@ -1,35 +0,0 @@
-apply plugin: 'com.jfrog.bintray'
```
markelliot, line 1: couldn't get this file to support multiple projects

markelliot:
@uschi2000

gradle/publish-libs.gradle
uschi2000, line 1: the nebula stuff you mean?

gradle/publish.gradle
```diff
@@ -0,0 +1,40 @@
...
+apply plugin: 'com.jfrog.bintray'
+
+task sourceJar(type: Jar) {
+    from sourceSets.main.allSource
+    classifier 'sources'
+}
+
+publishing {
+    publications {
+        bintray(MavenPublication) {
```
uschi2000, line 11: lost the javadoc jar. (which i personally don't care about too much, so i'm fine with it)

gradle/publish.gradle
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+bintray {
+    user = System.env.BINTRAY_USER
+    key = System.env.BINTRAY_KEY
+    publish = true
+    pkg {
+        repo = 'releases'
+        name = 'http-remoting'
```
uschi2000: meep

gradle.properties
```diff
@@ -2,10 +2,13 @@
 bintrayVersion=1.4
 guavaVersion=16.0.1
 immutablesVersion=2.1.8
-jacksonVersion=2.4.4
+jacksonVersion=2.6.3
```
uschi2000, line 5: this could mean a backcompat break for folks taking an update. it's probably fine, but we should note this in the release notes.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingContextFilter.java
```diff
@@ -0,0 +1,97 @@
...
+                        Request jettyRequest = (org.eclipse.jetty.server.Request) request;
+
+                        UserIdentity userIdentity = new DefaultUserIdentity(null,
+                                new UsernamePrincipal(jwt.getUnverifiedUserId()), new String[0]);
+
+                        jettyRequest.setAuthentication(new UserAuthentication("BEARER-TOKEN", userIdentity));
+                    }
+
+                } catch (Throwable t) {
+                    log.error("Unable to decode authorization header", t);
```
uschi2000: should be warn, I think

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingContextFilter.java
```diff
@@ -0,0 +1,97 @@
...
+                    MDC.put("userId", jwt.getUnverifiedUserId());
+
+                    // Jetty
+                    if (request instanceof org.eclipse.jetty.server.Request) {
+                        Request jettyRequest = (org.eclipse.jetty.server.Request) request;
+
+                        UserIdentity userIdentity = new DefaultUserIdentity(null,
+                                new UsernamePrincipal(jwt.getUnverifiedUserId()), new String[0]);
+
+                        jettyRequest.setAuthentication(new UserAuthentication("BEARER-TOKEN", userIdentity));
```
uschi2000, line 70: what's the effect of setting this on a jetty Request? is it just used for logging? feels a bit unintuitive ...

auth-tokens-filter/src/test/java/com/palantir/tokens/auth/http/BearerTokenLoggingContextFilterTests.java
```diff
@@ -0,0 +1,127 @@
...
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class BearerTokenLoggingContextFilterTests {
+
+    private static final String AUTH_HEADER = "Bearer eyJhbGciOiJFUzI1NiJ9."
```
uschi2000: can we replace this by something like:

```
AUTH_HEADER = "unused." + base64UrlEncode("{\"sub\": \"c393f659-0301-434e-a9c9-72304a507ffc\"") + ".unused";
```

? (just to make more obvious what's happening here)

uschi2000:
CR done

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingContextFilter.java
markelliot, line 70: Yeah, the goal is to get the userId injected into the Jetty request log (by default)...the test probably needs to show the string in the default log line not just that it shows up on the request log MDC...which it does but isn't asserted. Oops.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingContextFilter.java
markelliot: If you get here it means we weren't able to decode a provided bearer token. That seems worth investigating?

gradle.properties
markelliot, line 5: Had trouble getting Dropwizard to work with conflicting deps, this is the 0.9.2 dropwizard Jackson version. Can probably force this back down to what it was...will iterate.

gradle/publish-libs.gradle
markelliot, line 1: With old publish file, immediate Gradle error that says publishing does not support multiple projects. I'd seen the error before but in the interests of progress I just ripped the http-remoting Gradle file. 

We ought to do a cleanup pass on our open source project Gradle files -- they're all sight my drifted versions of each other due to evolution...would like to get back to a standard set.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingContextFilter.java
uschi2000: OK, if we hold ourselves to
https://github.com/palantir-baseline/best-practices/blob/master/java-coding-guidelines/readme.md#log-levels
,
then either error or warn would be fine, I guess.

auth-tokens-filter/src/test/java/com/palantir/tokens/auth/http/BearerTokenLoggingContextFilterTests.java
markelliot: done

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingContextFilter.java
markelliot: switched to warn based on my reading

gradle/publish.gradle
markelliot: fixed

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingContextFilter.java
markelliot, line 70: added an assert to prove this

markelliot:
couldn't get Dropwizard to run with old Jackson/started an internal thread on how we might generalize a solution here but it should be fine for consumers to use whatever version of Jackson they have as an override.

PR #14, opened 2016-07-12
BasicAuthToBearerTokenFilter

ags799:
WIP

### DIFF
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
@@ -35,7 +35,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletRequestWrapper;
 
 /**
- * A {@link Filter} that TODO(asharp).
+ * A {@link Filter} that replaces basic auth with a bearer token.
  */
 public class BasicAuthToBearerTokenFilter implements Filter {
     private static final String BASIC_AUTH_STR = "Basic";
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
@@ -78,7 +78,7 @@ public class BasicAuthToBearerTokenFilter implements Filter {
     }
 
     private ServletRequest getRequestWithTokenFromRawAuthHeader(String rawAuthHeader, HttpServletRequest request) {
-        Preconditions.checkNotNull(rawAuthHeader);
+        Preconditions.checkNotNull(rawAuthHeader, "null rawAuthHeader");
         if (isBasicAuth(rawAuthHeader)) {
             AuthHeader authHeader = base64DecodePassword(rawAuthHeader);
             return getRequestWithTokenFromAuthHeader(authHeader, request);
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
@@ -51,8 +51,10 @@ public class BasicAuthToBearerTokenFilter implements Filter {
         ServletRequest updatedRequest;
         try {
             updatedRequest = getRequestWithToken(request);
-        } catch (IllegalArgumentException e) {
-            log.warn(e.getMessage());
+        } catch (NullRawAuthHeaderException e) {
+            updatedRequest = request;
+        } catch (AuthHeaderNotBasicAuthException e) {
+            log.warn("AuthHeaderNotBasicAuthException", e);
             updatedRequest = request;
         }
         chain.doFilter(updatedRequest, response);
@@ -61,22 +63,26 @@ public class BasicAuthToBearerTokenFilter implements Filter {
     @Override
     public void destroy() {}
 
-    private ServletRequest getRequestWithToken(ServletRequest request) {
+    private ServletRequest getRequestWithToken(ServletRequest request)
+            throws AuthHeaderNotBasicAuthException, NullRawAuthHeaderException {
         if (request instanceof HttpServletRequest) {
             final HttpServletRequest httpRequest = (HttpServletRequest) request;
             final String rawAuthHeader = httpRequest.getHeader(HttpHeaders.AUTHORIZATION);
             return getRequestWithTokenFromRawAuthHeader(rawAuthHeader, httpRequest);
         } else {
-            throw new IllegalArgumentException("Request is not an HttpServletRequest.");
+            throw new AuthHeaderNotBasicAuthException("Request is not an HttpServletRequest.");
         }
     }
 
-    private ServletRequest getRequestWithTokenFromRawAuthHeader(String rawAuthHeader, HttpServletRequest request) {
-        if (rawAuthHeader != null && isBasicAuth(rawAuthHeader)) {
+    private ServletRequest getRequestWithTokenFromRawAuthHeader(String rawAuthHeader, HttpServletRequest request)
+            throws AuthHeaderNotBasicAuthException, NullRawAuthHeaderException {
+        if (rawAuthHeader == null) {
+            throw new NullRawAuthHeaderException("Raw auth header is null.");
+        } else if (isBasicAuth(rawAuthHeader)) {
             final AuthHeader authHeader = getAuthHeader(rawAuthHeader);
             return getRequestWithTokenFromAuthHeader(authHeader, request);
         } else {
-            throw new IllegalArgumentException("Auth header is not basic auth.");
+            throw new AuthHeaderNotBasicAuthException("Auth header is not basic auth.");
         }
     }
 
@@ -113,4 +119,15 @@ public class BasicAuthToBearerTokenFilter implements Filter {
         return rawAuthHeader.contains(BASIC_AUTH_STR);
     }
 
+    class NullRawAuthHeaderException extends Exception {
+        public NullRawAuthHeaderException(String message) {
+            super(message);
+        }
+    }
+
+    class AuthHeaderNotBasicAuthException extends Exception {
+        public AuthHeaderNotBasicAuthException(String message) {
+            super(message);
+        }
+    }
 }
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
@@ -48,35 +48,54 @@ public class BasicAuthToBearerTokenFilter implements Filter {
     @Override
     public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
             throws IOException, ServletException {
-        if (request instanceof HttpServletRequest) {
-            HttpServletRequest httpRequest = (HttpServletRequest) request;
-            String rawAuthHeader = httpRequest.getHeader(HttpHeaders.AUTHORIZATION);
-            if (rawAuthHeader != null && rawAuthHeader.contains(BASIC_AUTH_STR)) {
-                String password = getPassword(rawAuthHeader);
-                final AuthHeader authHeader;
-                try {
-                    authHeader = AuthHeader.valueOf(password);
-                } catch (IllegalArgumentException e) {
-                    throw new IllegalArgumentException(
-                            "Could not deserialize bearer token from password field: " + e.getMessage(), e);
-                }
-                HttpServletRequestWrapper wrapper = new HttpServletRequestWrapper(httpRequest) {
-                    @Override
-                    public String getHeader(String name) {
-                        if(Objects.equals(name, HttpHeaders.AUTHORIZATION)) {
-                            return authHeader.toString();
-                        } else {
-                            return super.getHeader(name);
-                        }
-                    }
-                };
-
-                chain.doFilter(wrapper, response);
-            } else {
-                log.warn("Authorization header does not contain basic auth sentinel string.");
-            }
+        ServletRequest updatedRequest;
+        try {
+            updatedRequest = getRequestWithToken(request);
+        } catch (IllegalArgumentException e) {
+            log.warn(e.getMessage());
+            updatedRequest = request;
         }
-        chain.doFilter(request, response);
+        chain.doFilter(updatedRequest, response);
+    }
+
+    @Override
+    public void destroy() {}
+
+    private ServletRequest getRequestWithToken(ServletRequest request) {
+        if (request instanceof HttpServletRequest) {
+            final HttpServletRequest httpRequest = (HttpServletRequest) request;
+            final String rawAuthHeader = httpRequest.getHeader(HttpHeaders.AUTHORIZATION);
+            return getRequestWithTokenFromRawAuthHeader(rawAuthHeader, httpRequest);
+        } else {
+            throw new IllegalArgumentException("Request is not an HttpServletRequest.");
+        }
+    }
+
+    private ServletRequest getRequestWithTokenFromRawAuthHeader(String rawAuthHeader, HttpServletRequest request) {
+        if (rawAuthHeader != null && isBasicAuth(rawAuthHeader)) {
+            final AuthHeader authHeader = getAuthHeader(rawAuthHeader);
+            return getRequestWithTokenFromAuthHeader(authHeader, request);
+        } else {
+            throw new IllegalArgumentException("Auth header is not basic auth.");
+        }
+    }
+
+    private ServletRequest getRequestWithTokenFromAuthHeader(final AuthHeader authHeader, HttpServletRequest request) {
+        return new HttpServletRequestWrapper(request) {
+            @Override
+            public String getHeader(String name) {
+                if(Objects.equals(name, HttpHeaders.AUTHORIZATION)) {
+                    return authHeader.toString();
+                } else {
+                    return super.getHeader(name);
+                }
+            }
+        };
+    }
+
+    private AuthHeader getAuthHeader(String rawAuthHeader) {
+        final String password = getPassword(rawAuthHeader);
+        return AuthHeader.valueOf(password);
     }
 
     private String getPassword(String rawAuthHeader) {
@@ -90,7 +109,8 @@ public class BasicAuthToBearerTokenFilter implements Filter {
         return credentials.split(":", 2)[1];
     }
 
-    @Override
-    public void destroy() {}
+    private boolean isBasicAuth(String rawAuthHeader) {
+        return rawAuthHeader.contains(BASIC_AUTH_STR);
+    }
 
 }
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
@@ -70,8 +70,8 @@ public class BasicAuthToBearerTokenFilter implements Filter {
     private ServletRequest getRequestWithToken(ServletRequest request)
             throws AuthHeaderNotBasicAuthException, NullRawAuthHeaderException {
         if (request instanceof HttpServletRequest) {
-            final HttpServletRequest httpRequest = (HttpServletRequest) request;
-            final String rawAuthHeader = httpRequest.getHeader(HttpHeaders.AUTHORIZATION);
+            HttpServletRequest httpRequest = (HttpServletRequest) request;
+            String rawAuthHeader = httpRequest.getHeader(HttpHeaders.AUTHORIZATION);
             return getRequestWithTokenFromRawAuthHeader(rawAuthHeader, httpRequest);
         } else {
             throw new AuthHeaderNotBasicAuthException("Request is not an HttpServletRequest.");
@@ -83,7 +83,7 @@ public class BasicAuthToBearerTokenFilter implements Filter {
         if (rawAuthHeader == null) {
             throw new NullRawAuthHeaderException("Raw auth header is null.");
         } else if (isBasicAuth(rawAuthHeader)) {
-            final AuthHeader authHeader = getAuthHeader(rawAuthHeader);
+            AuthHeader authHeader = getAuthHeader(rawAuthHeader);
             return getRequestWithTokenFromAuthHeader(authHeader, request);
         } else {
             throw new AuthHeaderNotBasicAuthException("Auth header is not basic auth.");
@@ -104,7 +104,7 @@ public class BasicAuthToBearerTokenFilter implements Filter {
     }
 
     private AuthHeader getAuthHeader(String rawAuthHeader) {
-        final String password = getPassword(rawAuthHeader);
+        String password = getPassword(rawAuthHeader);
         return AuthHeader.valueOf(password);
     }
 
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
@@ -81,7 +81,11 @@ public class BasicAuthToBearerTokenFilter implements Filter {
     }
 
     private ServletRequest getRequestWithTokenFromRawAuthHeader(String rawAuthHeader, HttpServletRequest request) {
-        Preconditions.checkNotNull(rawAuthHeader, "null rawAuthHeader");
+        try {
+            Preconditions.checkNotNull(rawAuthHeader, "null rawAuthHeader");
+        } catch (NullPointerException e) {
+            throw new IllegalArgumentException(e);
+        }
         if (isBasicAuth(rawAuthHeader)) {
             AuthHeader authHeader = base64DecodePassword(rawAuthHeader);
             return getRequestWithTokenFromAuthHeader(authHeader, request);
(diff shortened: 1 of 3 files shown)


jared2501:
Hey @ags799 can you fix the build issues? The circle build is failing on checkstyle. Also can you update the PR description?

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,133 @@
...
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+
+/**
+ * A {@link Filter} that replaces basic auth with a bearer token.
```
jared2501, line 39: any chance of a bit more of a description about how it does this? I.e. it assumes that the bearer token is in the password field of the basic auth header.

ags799, line 39: Done.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,135 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A {@link Filter} that replaces basic auth with a bearer token. It assumes the bearer token is in the password field
```
gracew: typically the first sentence should be its own paragraph. insert a `<p>` tag before subsequent paragraphs

gracew: can you also specify that the password field is expected to be the base 64 encoded version of the bearer token?

build.gradle
```diff
@@ -1,5 +1,6 @@
 buildscript {
     repositories {
+        mavenLocal()
```
gracew: delete

gracew:
@ags799  not sure off the top of my head. you may want to retrigger the build with SSH so you can inspect the failure more closely. it would also be good to contribute a PR that copies the test report back as a circle artifact so that failures are more easily inspected in the future.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public void destroy() {
+    }
+
+    private ServletRequest getRequestWithToken(ServletRequest request)
+            throws AuthHeaderNotBasicAuthException, NullRawAuthHeaderException {
+        if (request instanceof HttpServletRequest) {
+            final HttpServletRequest httpRequest = (HttpServletRequest) request;
```
gracew: no need to specify local variables as final
http://stackoverflow.com/questions/18225572/use-of-final-local-variables-in-java

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,135 @@
...
+            }
+        };
+    }
+
+    private AuthHeader getAuthHeader(String rawAuthHeader) {
+        final String password = getPassword(rawAuthHeader);
+        return AuthHeader.valueOf(password);
+    }
+
+    private String getPassword(String rawAuthHeader) {
```
gracew: not sure it's valuable to have so many helper methods -- for example, I think this one can be inlined with `getAuthHeader`

build.gradle
ags799: Done.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,135 @@
...
+                if (Objects.equals(name, HttpHeaders.AUTHORIZATION)) {
+                    return authHeader.toString();
+                } else {
+                    return super.getHeader(name);
+                }
+            }
+        };
+    }
+
+    private AuthHeader getAuthHeader(String rawAuthHeader) {
```
gracew: I think a more descriptive method name might be something like `base64DecodePassword`

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,135 @@
...
+            return getRequestWithTokenFromRawAuthHeader(rawAuthHeader, httpRequest);
+        } else {
+            throw new AuthHeaderNotBasicAuthException("Request is not an HttpServletRequest.");
+        }
+    }
+
+    private ServletRequest getRequestWithTokenFromRawAuthHeader(String rawAuthHeader, HttpServletRequest request)
+            throws AuthHeaderNotBasicAuthException, NullRawAuthHeaderException {
+        if (rawAuthHeader == null) {
+            throw new NullRawAuthHeaderException("Raw auth header is null.");
```
gracew: I don't think we need to define more specific exceptions. using guava's preconditions (https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Preconditions.html) should be sufficient. For example:

```
Preconditions.checkNotNull(rawAuthHeader);
```

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,135 @@
...
+    private static final BaseEncoding BASE_64_ENCODING = BaseEncoding.base64Url();
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+    }
+
+    @Override
+    public final void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        ServletRequest updatedRequest;
```
gracew: might make sense to default the value of `updatedRequest` to `request`

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,135 @@
...
+            throw new NullRawAuthHeaderException("Raw auth header is null.");
+        } else if (isBasicAuth(rawAuthHeader)) {
+            final AuthHeader authHeader = getAuthHeader(rawAuthHeader);
+            return getRequestWithTokenFromAuthHeader(authHeader, request);
+        } else {
+            throw new AuthHeaderNotBasicAuthException("Auth header is not basic auth.");
+        }
+    }
+
+    private ServletRequest getRequestWithTokenFromAuthHeader(final AuthHeader authHeader, HttpServletRequest request) {
```
gracew: think this method can be combined with the previous one

auth-tokens-filter/src/test/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilterTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertRequestHasAuthHeader("Bearer " + password);
+    }
+
+    private void setPassword(String password) {
+        final String authHeader = "Basic " + base64Encode("foo:" + password);
+        when(request.getHeader(HttpHeaders.AUTHORIZATION)).thenReturn(authHeader);
+    }
+
+    private void doFilter() throws IOException, ServletException {
+        tokenFilter.doFilter(request, null, chain);
```
gracew, line 112: as a one line method I think this can just be inlined in the test, especially because this is the main focus of the test.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: Done.

auth-tokens-filter/src/test/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilterTest.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private void assertRequestHasAuthHeader(String expectedAuthHeader) throws IOException, ServletException {
+        verify(chain).doFilter(requestArgumentCaptor.capture(), Mockito.<ServletResponse>any());
+        final HttpServletRequest value = requestArgumentCaptor.getValue();
+        final String actualAuthHeader = value.getHeader(HttpHeaders.AUTHORIZATION);
+        MatcherAssert.assertThat(actualAuthHeader, Is.is(expectedAuthHeader));
+    }
+
+    private String base64Encode(String str) {
```
gracew: inline with `setPassword`

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: Done.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private String getPassword(String rawAuthHeader) {
+        String base64Credentials = rawAuthHeader.substring(BASIC_AUTH_STR.length()).trim();
+        String credentials;
+        try {
+            credentials = new String(BASE_64_ENCODING.decode(base64Credentials), StandardCharsets.UTF_8);
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException("Could not decode credentials from auth header: " + e.getMessage());
+        }
+        return credentials.split(":", 2)[1];
```
gracew: we should make sure there's actually a colon in the string and throw an exception if not -- otherwise this could result in a nasty `IndexOutOfBoundsException`

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: I'm happy to inline `isBasicAuth`, but I think `getPassword` is large enough to warrant its own method.

NVM I see how to do this. Done.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: Done.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: Done.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: IMO it'd be too much nesting.

auth-tokens-filter/src/test/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilterTest.java
ags799, line 112: Done.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: Yes would rather not have custom exceptions. The trouble is we want to respond differently to different failures: if the `rawAuthHeader` is null, we don't want to log anything. If it's not null, and it's not in basic auth format, we want to log that.

Commit `b5d03a` removes the custom exceptions. The penalty is we have some code duplication with logging an error message and then throwing an `IllegalArgumentException` with that message.

auth-tokens-filter/src/test/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilterTest.java
ags799: Done.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: Good idea. Done.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * A {@link Filter} that replaces basic auth with a bearer token.
+ *
+ * <p>It assumes that the bearer token is held in the password field of the basic auth credentials, and that it is
+ * base-64 encoded.
+ */
+public class BasicAuthToBearerTokenFilter implements Filter {
+    private static final String BASIC_AUTH_STR = "Basic";
+    private static final Logger log = LoggerFactory.getLogger(BasicAuthToBearerTokenFilter.class);
```
gracew, line 45: (nit) I prefer to put the logger in its own section b/c it's sort of a meta variable. like this:

```
private static final Logger log = ...;

private static final String BASIC_AUTH_STR = ...
private static final BaseEncoding BASE_64_ENCODING = ...
```

ags799, line 45: Done.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,128 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A {@link Filter} that replaces basic auth with a bearer token.
+ *
+ * <p>It assumes that the bearer token is held in the password field of the basic auth credentials, and that it is
+ * base-64 encoded.
+ */
+public class BasicAuthToBearerTokenFilter implements Filter {
```
qinfchen, line 44: Does it make sense to have a single filter that checks for either "Basic" or  "Bearer"?

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,129 @@
...
+            return getRequestWithTokenFromRawAuthHeader(rawAuthHeader, httpRequest);
+        } else {
+            String message = "Request is not an HttpServletRequest.";
+            log.warn(message);
+            throw new IllegalArgumentException(message);
+        }
+    }
+
+    private ServletRequest getRequestWithTokenFromRawAuthHeader(String rawAuthHeader, HttpServletRequest request) {
+        Preconditions.checkNotNull(rawAuthHeader, "null rawAuthHeader");
```
qinfchen: this is private method, no need to check for null?

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,129 @@
...
+            log.warn(message);
+            throw new IllegalArgumentException(message);
+        }
+    }
+
+    private ServletRequest getRequestWithTokenFromAuthHeader(final AuthHeader authHeader, HttpServletRequest request) {
+        return new HttpServletRequestWrapper(request) {
+            @Override
+            public String getHeader(String name) {
+                if (Objects.equals(name, HttpHeaders.AUTHORIZATION)) {
```
qinfchen, line 96: Strings.equals? object.equals compares memory reference, right?

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: `rawAuthHeader` may be null, and there is no check prior to this method's invocation.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799, line 44: @jared2501 ?

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
qinfchen: ahh, nevermind.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799, line 96: Using `Objects` not `Object`.

https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#equals(java.lang.Object,%20java.lang.Object

pnepywoda:
you know, I wonder if this should instead go into http-remoting. we already have a dependency on [auth-tokens](https://github.com/palantir/http-remoting/blob/develop/service-config/build.gradle#L8) there

pnepywoda:
or, we could make a different sub-project here in auth-tokens named maybe "http-servers" which contains the filter and can then depend on javax.ws.rs to use things like NotAuthorizedException

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,129 @@
...
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A {@link Filter} that replaces basic auth with a bearer token.
+ *
+ * <p>It assumes that the bearer token is held in the password field of the basic auth credentials. It assumes that the
```
gracew: the format should be:

```
/**
 * first sentence.
 * <p>
 * second sentence. third sentence. etc
 */
```

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,129 @@
...
+        String credentials;
+        try {
+            credentials = new String(BASE_64_ENCODING.decode(base64Credentials), StandardCharsets.UTF_8);
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException("Could not decode credentials from auth header: " + e.getMessage());
+        }
+        String password;
+        try {
+            password = credentials.split(":", 2)[1];
+        } catch (IndexOutOfBoundsException e) {
```
gracew: it's not great to use exceptions for control flow. can we instead do a `checkArgument(credentials.contains(":"))`

auth-tokens-filter/src/test/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilterTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.hamcrest.core.Is;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+public class BasicAuthToBearerTokenFilterTest {
```
gracew, line 42: let's add some more tests that check:
- non HttpServletRequests
- non-decodable basic auth header
- basic auth header without a colon
- etc

gracew:
ok, I think this is close after the addition of some more unit tests. @pnepywoda , I think a separate project makes sense. @jared2501 and I also wanted to modify this lib to throw a 401 for null auth headers which would also require a javax.ws.rs dependency.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: Done.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: Done.

auth-tokens-filter/src/test/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilterTest.java
ags799, line 42: Done.

gracew:
+1, would like to merge this

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,129 @@
...
+            log.warn(message);
+            throw new IllegalArgumentException(message);
+        }
+    }
+
+    private ServletRequest getRequestWithTokenFromRawAuthHeader(String rawAuthHeader, HttpServletRequest request) {
+        try {
+            Preconditions.checkNotNull(rawAuthHeader, "null rawAuthHeader");
+        } catch (NullPointerException e) {
+            throw new IllegalArgumentException(e);
```
markelliot: I think you're trying to throw an `IllegalArgumentException` based on a null check; either convert this code to directly check nullity, or replace the try/catch with:

```
Preconditions.checkArgument(rawAuthHeader != null, "null rawAuthHeader is not allowed");
```

...but let's not use exceptions for control flow.

ags799: Thanks, I redesigned so that exceptions aren't used for control flow. Let me know what you think.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    private ServletRequest addBearerTokenIfBasicAuth(HttpServletRequest request, String rawAuthHeader) {
+        if (isBasicAuth(rawAuthHeader)) {
+            AuthHeader authHeader;
+            try {
+                authHeader = base64DecodePassword(rawAuthHeader);
+            } catch (IllegalArgumentException e) {
+                log.warn("IllegalArgumentException: " + e.getMessage());
```
markelliot: probably want to say "Could not decode password in basic auth header". Also, since you have the exception, can just log the actual exception:

```
log.warn("Could not decode password in basic auth header", e);
```

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -0,0 +1,122 @@
...
+                return request;
+            }
+            return addBearerTokenIfBasicAuth(request, authHeader);
+        } else {
+            log.warn("Auth header is not basic auth.");
+            return request;
+        }
+    }
+
+    private ServletRequest addBearerTokenIfBasicAuth(HttpServletRequest request, final AuthHeader authHeader) {
```
markelliot: Method name is `...IfBasicAuth`, but actually this runs always.

markelliot:
thanks, couple of nits but looks great

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: Done.

ags799:
FYI, the last failure comes from a class that this PR does not touch. The failure has occurred on this PR before, it appears flaky. It's `com.palantir.tokens.auth.http.BearerTokenLoggingContextFilterTests > testBearerTokenLogging`

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
ags799: Good idea, I've removed `IfBasicAuth` from method name.

markelliot:
This was released as part of 0.2.6, which should now be available.

I took a stab at addressing the test flake seen here in #18, though I'm not sure it's truly fixed. The thing its testing is particularly squirrelly because of how nested it is inside Dropwizard and success depends on an async call in a location that's hard to latch -- I tweaked some setup and run the request twice to up the chances the order happens right, though obviously this is less than ideal.

PR #15, opened 2016-07-14
copy artifacts to circle

ags799:
make it easier to review build errors

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,8 +35,7 @@ allprojects {
             outputs.upToDateWhen { false }
             doLast {
                 [
-                        file("auth-tokens/build/reports"),
-                        file("auth-tokens-filter/build/reports")
+                        file("${buildDir}")
                 ].each { src ->
                     copy {
                         from src
--- a/build.gradle
+++ b/build.gradle
@@ -29,6 +29,25 @@ allprojects {
     }
 
     version gitVersion()
+
+    if (System.env.CIRCLE_ARTIFACTS) {
+        task copyCiArtifactsToCircle {
+            outputs.upToDateWhen { false }
+            doLast {
+                [
+                        file("${projectDir}/contiperf-report"), // contiperf benchmark results
+                        file("${buildDir}/reports"), // jacoco, checkstyle, findbugs
+                        file("${buildDir}/docs/javadoc/"),
+                        file("${projectDir}/var/log") // docker container logs
+                ].each { src ->
+                    copy {
+                        from src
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -29,6 +29,23 @@ allprojects {
     }
 
     version gitVersion()
+
+    if (System.env.CIRCLE_ARTIFACTS) {
+        task copyCiArtifactsToCircle {
+            outputs.upToDateWhen { false }
+            doLast {
+                [
+                        file("auth-tokens/build/reports"),
```
gracew: we shouldn't hardcode this -- let's use `${buildDir}` (that will apply to all projects and we won't have to have two lines here)

build.gradle
```diff
@@ -29,6 +29,23 @@ allprojects {
     }
 
     version gitVersion()
+
+    if (System.env.CIRCLE_ARTIFACTS) {
+        task copyCiArtifactsToCircle {
```
gracew, line 34: we need to call this task somewhere in circle.yml

ags799, line 34: Should I see `copyCiArtifactsToCircle` in `./gradlew tasks`?

build.gradle
ags799: Done.

build.gradle
jared2501, line 34: checkout https://github.palantir.build/elements/stemma/blob/develop/circle.yml#L52

We call it in the post test block in circle so that circle runs it after all the things

gracew:
sweet! I see an Artifacts tab now :)

@pnepywoda , @qinfchen can we merge?

markelliot:
can we just set the buildDir to be CIRCLE_ARTIFACTS rather than have this copy task?

markelliot:
scratch that, can we just do this the Circle CI way and add a general/artifacts block like in https://github.com/palantir/encrypted-config-value/blob/develop/circle.yml#L9-L12

markelliot:
...no wait, we already have that too. What is this solving?

ags799:
@gracew where can you see the artifacts tab?

gracew:
@markelliot the problem this is solving is copying back the junit test report so that we can see it in the Circle CI build page.

@ags799 I see it here: https://circleci.com/gh/palantir/auth-tokens/72

screen shot 2016-07-20 at 10 42 16 pm

markelliot:
Cool, #17 collects test artifacts the standard way into the Circle test directory so that we get the summary and pretty-printed alerts. Closing this since I just merged #17.

PR #18, opened 2016-07-21
Resolve test flake

markelliot:


markelliot:
Ran through a few iterations, pretty sure there's a race between when the request goes through the request log pipeline and when we ask for the result of the request. Because of the plumbing, this is pretty gnarly to fix, can come back to it again with something more rigorous if it continues to flake.

Issue #21, opened 2016-08-09
Please add AuthHeader serializability note to 0.2.4 release notes


ash211:
At https://github.com/palantir/auth-tokens/releases/tag/0.2.4

Proposed language:
- Note that `AuthHeader` is [no longer](https://github.com/palantir/auth-tokens/commit/a8a475aea076ce77eda546af917207967f5d08f0) JSON-serializable as it should be used for `@HeaderParam`s only.  Use `BearerToken` instead which is and should be serializable.

This is technically a backcompat breaking change (I had to change a downstream project to accommodate) so we should at least make note of it in release notes even though we're still in 0.x versions.

cc @pnepywoda

EthanLozano:
Alternatively, could we just revert the change? https://github.com/palantir/auth-tokens/pull/22

pnepywoda:
I don't have edit permissions on the release notes, @markelliot can you make me an admin of this project?

markelliot:
I'm not an administrator of this repo but I updated the release notes.

ash211:
Thanks Mark.  Closing.

I don't see a strong reason to not have AuthHeaders be Jackson-serializable, but will let that discussion happen on another ticket.

PR #22, opened 2016-08-30
Revert "remove json serialization annotations"

EthanLozano:
This reverts commit a8a475aea076ce77eda546af917207967f5d08f0.

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
@@ -16,6 +16,8 @@
 
 package com.palantir.tokens.auth;
 
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import org.immutables.value.Value;
 
 /**
@@ -23,8 +25,8 @@ import org.immutables.value.Value;
  */
 @Value.Immutable
 @Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE, jdkOnly = true)
-// NOTE: no @JsonSerialize/@JsonDeserialize because auth headers are for use in @HeaderParam
-// see: https://jersey.java.net/apidocs/latest/jersey/javax/ws/rs/HeaderParam.html
+@JsonSerialize(as = ImmutableAuthHeader.class)
+@JsonDeserialize(as = ImmutableAuthHeader.class)
 public abstract class AuthHeader {
 
     @Value.Parameter


EthanLozano:
The removal of the serializer and deserializer broke my code. Specifically, I rely on the ability to serialize and deserialize an AuthHeader (in addition to using it in the header). Also, the ImmutableAuthHeader is package-private, so I don't have access to the auto-generated implementation. Would it be possible to add this code back?

qinfchen:
`ImmutableAuthHeader` should be package-private. Why not just `AuthHeader`'s static methods to build an instance of `AuthHeader`?

EthanLozano:
To clarify, I'm asking to bring the `@JsonSerialize` and `@JsonDeserialize` annotations back. Otherwise, it looks like I need to write a custom serializer, a custom deserializer, and register both with my object mappers during serialization and deserialization.

qinfchen:
deferring that to @pnepywoda.

EthanLozano:
As far as I know, the code below is the minimal amount of code that I need to regain the serialization and deserialization functionality for AuthHeader. Let me know if there is a more succinct way to regain the functionality. Alternatively, if we added the serialization/deserialization functionality back, then this extra code wouldn't be necessary.

CustomSerializableObject.java:

``` java
    @JsonSerialize(using = AuthHeaderSerializer.class)
    @JsonDeserialize(using = AuthHeaderDeserializer.class)
    private final AuthHeader authHeader;
```

AuthHeaderSerializer.java:

``` java
import com.fasterxml.jackson.core.JsonGenerator;
import com.fasterxml.jackson.databind.SerializerProvider;
import com.fasterxml.jackson.databind.ser.std.StdSerializer;
import com.palantir.tokens.auth.AuthHeader;
import java.io.IOException;

class AuthHeaderSerializer extends StdSerializer<AuthHeader> {

    public static final String BEARER_KEY = "Bearer";

    public AuthHeaderSerializer() {
        this(null);
    }

    public AuthHeaderSerializer(Class<AuthHeader> itemClass) {
        super(itemClass);
    }

    @Override
    public void serialize(AuthHeader value, JsonGenerator gen, SerializerProvider provider) throws IOException {
        gen.writeStartObject();
        gen.writeStringField(BEARER_KEY, value.getBearerToken().getToken());
        gen.writeEndObject();
    }

}
```

AuthHeaderDeserializer.java:

``` java
import com.fasterxml.jackson.core.JsonParser;
import com.fasterxml.jackson.core.JsonProcessingException;
import com.fasterxml.jackson.databind.DeserializationContext;
import com.fasterxml.jackson.databind.JsonNode;
import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
import com.palantir.tokens.auth.AuthHeader;
import com.palantir.tokens.auth.BearerToken;
import java.io.IOException;

class AuthHeaderDeserializer extends StdDeserializer<AuthHeader> {

    public AuthHeaderDeserializer() {
        this(null);
    }

    public AuthHeaderDeserializer(Class<?> vc) {
        super(vc);
    }

    @Override
    public AuthHeader deserialize(JsonParser jsonParser, DeserializationContext ctxt)
        throws IOException, JsonProcessingException {
        JsonNode node = jsonParser.getCodec().readTree(jsonParser);
        String bearerTokenStr = node.get(AuthHeaderSerializer.BEARER_KEY).asText();
        BearerToken bearerToken = BearerToken.valueOf(bearerTokenStr);
        return AuthHeader.of(bearerToken);
    }

}
```

j-baker:
Hey Ethan - jackson mixins are generally good for this sort of stuff if you really need them, and save ~= all the pain you suffer from.
I guess I don't understand why you need to jsonserialise an authheader and not just the bearer token itself?
James

pnepywoda:
I'll add the breaking change to the release notes. AuthHeader is for header information only and shouldn't ever be jackson serialized - if you need to serialize the token, read out the BearerToken

EthanLozano:
Thanks for suggesting mixins, that is a nice alternative. I agree that serializing the bearer token is simple, but the bigger problem is that this is a backwards compatibility breaking change ( https://github.com/palantir/auth-tokens/issues/21 ). I guess I don't understand why the AuthHeader can't both be used in a header and serialized

pnepywoda:
It's because it's fundamentally a _HEADER_ (it's for use in http headers) and NOT for serialization in a http BODY.

markelliot:
Why aren't we just using `@JsonValue` and `@JsonCreator` here?

markelliot:
AuthHeader is fundamentally serializable, it's just that it serializes to and from a string, not a richer json object.

pnepywoda:
Like I said, AuthHeader is for headers and doesn't/shouldn't go through json ser-de

PR #24, opened 2016-10-05
Extract unverified session id from token

markelliot:


markelliot:
@uschi2000 pulling this out based on requests to tie sessions together -- most of the interesting traffic to leaf nodes should have a valid session id in it. Also, I wonder if it's worth considering adding a `VerifiedSessionToken` that takes a public key (which we can fetch via, say, discovery).

gradle.properties
```diff
@@ -23,4 +23,4 @@ nebulaPublishingVersion=4.4.4
 
 # Processors
 immutablesVersion=2.1.15
-processorVersion=1.2.1
+processorVersion=1.2.3
```
markelliot, line 26: required to get this project working in intellij =/

uschi2000:
👍

PR #33, opened 2017-05-03
Add ImmutablesStyle annotation to remove warnings

### LINKED ISSUE #28: Immutable Style classes not found
The `auth-tokens` module uses `@Immutables` but does not include the annotations as compile time dependencies (it cannot find the styles classes). This causes downstream java warnings.

Two possible solutions would be to:

1. include the annotations in the classpath:

    ```
    compile group: 'org.immutables', name: 'value', classifier: 'annotations'
    ```

2. create and switch to an annotation provided in the `auth-tokens` module itself (i.e. something named `@ImmutableStyles`)

would a PR that does 2 make sense?

mnazbro:
Fixes #28

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,9 +4,6 @@ buildscript {
         maven {
             url 'https://plugins.gradle.org/m2/'
         }
-        maven {
-            url  "http://palantir.bintray.com/releases"
-        }
     }
 
     dependencies {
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -4,6 +4,9 @@ buildscript {
         maven {
             url 'https://plugins.gradle.org/m2/'
         }
+        maven {
+            url  "http://palantir.bintray.com/releases"
```
markelliot: why is this required?

markelliot:
isn't #28 still a problem because this uses `@Value.Immutable`?

build.gradle
mnazbro: I was actually having trouble with building this locally without adding bintray, I'll try to remove it and see if it isn't.

mnazbro:
This should fix the problem. The answer given from the Immutables authors: https://github.com/immutables/immutables/issues/291#issuecomment-213574619

build.gradle
mnazbro: I fixed this by changing the group in build.gradle instead.

PR #30, opened 2017-03-24
Bump Gradle to 3.4.1 and make non-API dependencies implementation

nathklei:
This removes Guava from the compile classpath of dependent libraries, andkeeps it on Runtime, unless explicitly added.

This is not back compatible if a library depended on inheriting the compile dependency on these libraries.

PR #35, opened 2017-05-28
Dholanda/remove guava dependencies

diogoholanda:
Removing guava dependencies from the project.
As a consequence, the project was bumped from java 7 to java 8.
All tests are still passing, so the JSON serialization seems to have been kept intact.

The Preconditions class is almost an exact copy from guava's. Are there any license requirements we need to comply with to do so? This alternative seemed easier than shading, since we use a single method from the whole package.

The second commit could arguably be a separate PR.

### DIFF
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth2/http/BearerTokenLoggingFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth2/http/BearerTokenLoggingFilter.java
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-package com.palantir.tokens2.auth.http;
+package com.palantir.tokens.auth2.http;
 
-import com.palantir.tokens2.auth.AuthHeader;
-import com.palantir.tokens2.auth.UnverifiedJsonWebToken;
+import com.palantir.tokens.auth2.AuthHeader;
+import com.palantir.tokens.auth2.UnverifiedJsonWebToken;
 import java.util.Optional;
 import javax.annotation.Priority;
 import javax.ws.rs.Priorities;
@@ -79,6 +79,6 @@ public class BearerTokenLoggingFilter implements ContainerRequestFilter {
     }
 
     public static String getRequestPropertyKey(String key) {
-        return "com.palantir.tokens2.auth." + key;
+        return "com.palantir.tokens.auth2." + key;
     }
 }
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BasicAuthToBearerTokenFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BasicAuthToBearerTokenFilter.java
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-package com.palantir.tokens.auth.http;
+package com.palantir.tokens2.auth.http;
 
-import com.palantir.tokens.auth.AuthHeader;
-import com.palantir.tokens.auth.Preconditions;
+import com.palantir.tokens2.auth.AuthHeader;
+import com.palantir.tokens2.auth.Preconditions;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BasicAuthToBearerTokenFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BasicAuthToBearerTokenFilter.java
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-package com.palantir.tokens.auth2.http;
+package com.palantir.tokens2.auth.http;
 
-import com.palantir.tokens.auth2.Preconditions;
-import com.palantir.tokens.auth2.AuthHeader;
+import com.palantir.tokens2.auth.Preconditions;
+import com.palantir.tokens2.auth.AuthHeader;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BearerTokenLoggingFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BearerTokenLoggingFilter.java
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-package com.palantir.tokens.auth2.http;
+package com.palantir.tokens2.auth.http;
 
-import com.palantir.tokens.auth2.AuthHeader;
-import com.palantir.tokens.auth2.UnverifiedJsonWebToken;
+import com.palantir.tokens2.auth.AuthHeader;
+import com.palantir.tokens2.auth.UnverifiedJsonWebToken;
 import java.util.Optional;
 import javax.annotation.Priority;
 import javax.ws.rs.Priorities;
@@ -79,6 +79,6 @@ public class BearerTokenLoggingFilter implements ContainerRequestFilter {
     }
 
     public static String getRequestPropertyKey(String key) {
-        return "com.palantir.tokens.auth2." + key;
+        return "com.palantir.tokens2.auth." + key;
     }
 }
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BearerTokenLoggingFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BearerTokenLoggingFilter.java
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-package com.palantir.tokens.auth.http;
+package com.palantir.tokens2.auth.http;
 
-import com.palantir.tokens.auth.AuthHeader;
-import com.palantir.tokens.auth.UnverifiedJsonWebToken;
+import com.palantir.tokens2.auth.AuthHeader;
+import com.palantir.tokens2.auth.UnverifiedJsonWebToken;
 import java.util.Optional;
 import javax.annotation.Priority;
 import javax.ws.rs.Priorities;
@@ -79,6 +79,6 @@ public class BearerTokenLoggingFilter implements ContainerRequestFilter {
     }
 
     public static String getRequestPropertyKey(String key) {
-        return "com.palantir.tokens.auth." + key;
+        return "com.palantir.tokens2.auth." + key;
     }
 }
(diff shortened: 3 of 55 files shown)


auth-tokens/src/main/java/com/palantir/tokens/auth/BearerTokens.java
```diff
@@ -78,7 +78,7 @@ public static BearerToken fromPath(Path path) throws IOException {
         }
 
         LOGGER.warn("No token file found in any of the configured paths: {}.", paths);
-        return Optional.absent();
+        return Optional.empty();
```
markelliot, line 81: this ends up being a major break?

markelliot:
license requirements: yeah, we need to be sure to include the Guava license header underneath our own, and by convention we include a line like "Derived from <url>". If there are subtle changes (not the case here) we prefer to include a description of the changes.

I think this causes a major break, should likely generate a package shift and a move to 'auth-tokens2'.

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerTokens.java
NeilRickards: Most projects have been switching to Java `Optional` without a major version bump, on the basis it's a Java client break but not a wire break.

markelliot: for service versions, sure. For libraries, especially one like this that's involved in diamond dependencies, we've been using the package numbering scheme.

auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BearerTokenLoggingFilter.java
```diff
@@ -79,6 +79,6 @@ private void setUnverifiedContext(ContainerRequestContext requestContext, String
     }
 
     public static String getRequestPropertyKey(String key) {
-        return "com.palantir.tokens.auth." + key;
+        return "com.palantir.tokens2.auth." + key;
```
diogoholanda, line 82: This would have to be accounted for when upgrading consumers of the MDC.
I could leave it unchanged, but it seems a worse option

diogoholanda:
@markelliot I've applied the suggested changes

auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -14,15 +14,13 @@
  * limitations under the License.
  */
 
-package com.palantir.tokens.auth.http;
+package com.palantir.tokens2.auth.http;
```
markelliot, line 17: sorry, I think we should be moving to `com.palantir.tokens.auth2`

auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BearerTokenLoggingFilter.java
markelliot: yeah, could go either way on this, probably would've left it unchanged, but as long as the filters are updated it's fine

auth-tokens/src/main/java/com/palantir/tokens2/auth/UnverifiedJsonWebToken.java
```diff
@@ -43,14 +41,7 @@
 public abstract class UnverifiedJsonWebToken {
```
markelliot, line 41: before we cut a release of 2.0, we may want to consider the class hierarchy around UnverifiedJsonWebToken and in particular whether we want to also support a VerifiedJsonWebToken

markelliot (commented):
nit around the package name.

@uschi2000 can you double check what we're doing here?

auth-tokens-filter/src/main/java/com/palantir/tokens2/auth/http/BearerTokenLoggingFilter.java
diogoholanda: http-remoting chose the approach of changing, so I decided to take the same approach

auth-tokens/src/main/java/com/palantir/tokens2/auth/UnverifiedJsonWebToken.java
diogoholanda: I am not sure what exactly do you have in mind for this class. My main concern is that it makes some assumptions on the format of the Bearer token. We can discuss this as part of a separate PR/issue.

auth-tokens/src/main/java/com/palantir/tokens/auth2/Preconditions.java
```diff
@@ -0,0 +1,80 @@
...
+ * Only a subset of the methods were copied.
+ * Small changes were made to respect this repository's code conventions.
+ * No change in behavior are expected from such changes.
+ */
+public final class Preconditions {
+    private Preconditions() {}
+
+    public static void checkArgument(boolean expression, String errorMessageTemplate, Object... errorMessageArgs) {
+        if (!expression) {
+            throw new IllegalArgumentException(format(errorMessageTemplate, errorMessageArgs));
```
qinfchen: can't you use java formatter?

diogoholanda: I don't want to silently/slightly change the Preconditions behavior. I don't see the value of reimplementing this and making sure nothing changes.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth2/http/BearerTokenLoggingFilter.java
```diff
@@ -79,6 +79,6 @@ private void setUnverifiedContext(ContainerRequestContext requestContext, String
     }
 
     public static String getRequestPropertyKey(String key) {
-        return "com.palantir.tokens.auth." + key;
+        return "com.palantir.tokens.auth2." + key;
```
uschi2000: hmm, this is tricky. on the fence if this should be changed or not. what was your rationale for changing it?

uschi2000 (changes requested):
- the package name change should be accompanied by a maven group change to `com.palantir.tokens2` in  https://github.com/palantir/auth-tokens/pull/35/files#diff-c197962302397baf3a4cc36463dce5eaR28
- that makes me wonder if java package `com.palantir.tokens2.auth` is more appropriate than `com.palantir.tokens.auth`.
- general nit: this PR conflates a lot of different things: package name change, remove guava, remove hamcrest. fine for now, per generally easier to discuss these independently.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth2/http/BearerTokenLoggingFilter.java
diogoholanda: I was in doubt, so I checked what was done in http-remoting: https://github.com/palantir/http-remoting/commit/689b315eebda9183a89ef84895fa20e133e9f484#diff-73656c4e1289f899124f962ba82c494f

diogoholanda:
@markelliot what do you think about the package names. I think I agree with @uschi2000 and prefer what I had originally.

markelliot:
sure, I guess since the publication group is `com.palantir.tokens` in build.gradle...speaking of which, I think that also needs to change? (not sure what that does to jcenter publications)

there's a different question, in that case, of why we have the `.auth` in the package name at all, anyway.

uschi2000:
yes, the maven group has to change, see my comment above.

diogoholanda:
Changed the package back. Hyphens are not allowed in package names and I preferred to do the smallest change. I've also changed the maven group. Before releasing 2.0, the README should be updated to point to the new coordinates

uschi2000 (approved):

uschi2000:
cheers, Diogo!

auth-tokens/src/main/java/com/palantir/tokens2/auth/Preconditions.java
```diff
@@ -0,0 +1,80 @@
...
+package com.palantir.tokens2.auth;
+
+/*
+ * Derived from com.google.common.base.Preconditions
+ *
+ * Only a subset of the methods were copied.
+ * Small changes were made to respect this repository's code conventions.
+ * No change in behavior are expected from such changes.
+ */
+public final class Preconditions {
```
schlosna, line 40: Did we intend to make this public as opposed to package private? If we leave this public, we probably want to add baseline checkstyle to avoid accidental imports of this in other projects.

diogoholanda: I did not make it package private because this class was actually used in 2 different packages. One alternative could be to move this class to its own project and to use the api/implementation dependencies in the newish java-library gradle plugin. I've never used such plugin before though, so I might be missing something

diogoholanda: Just noticed that this has been open for a while: https://github.com/palantir/auth-tokens/pull/30/files

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerTokens.java
pnepywoda: I'm *strongly* against making a `tokens2` change just for this one method. I'd estimate that it'll take 30 hours over the course of months for projects to update to tokens2 and in the meanwhile we'll be converting between the 2 AuthHeaders everywhere. I'd even estimate somewhere in the 10's of developer hours wasted on dealing with conversions.

Can we just take the small backcompat-break hit, change this from guava to java8, and REMOVE the `tokens2` package alltogether?

pnepywoda: https://github.com/palantir/auth-tokens/pull/48

j-baker: Yeah please, can we just kill the tokens2 package and move back to tokens? You end up converting between them _everywhere_ and it's a giant pain. It's especially galling since the only two classes most people see have no actual breaking changes, and the ones that did break have no real diamond dependency issues.

j-baker: another idea is to just not remove the guava dependency. Optional has been in Guava since 10.0 and there are no plans to even _deprecate_ guava optional 'for the foreseeable future'.

Given that the caller who wishes a Java optional can just use Guava 21+ and just call .toJavaOptional(), we might as well avoid the break entirely.

meditativeape: +1 to the proposal of moving back to `tokens` package. I've spent a significant portion of my time on the conversion between v1 and v2 tokens while migrating a service to remoting3, and I'd imagine every team will have to do this until all services are migrated to remoting3.

PR #42, opened 2017-06-08
Unverified token properties no longer carry an asterisk suffix "*"

uschi2000:


uschi2000:
@markelliot

uschi2000:
@gracew

gracew (approved):
the change looks fine. where did the asterix suffix come from originally -- a palantir convention or a broader one?

uschi2000:
mark's & my first attempt at this. thought it might make more clear that this is not the verbatim token.

PR #48, opened 2017-08-17
[BREAK] Rename "tokens2" back to "tokens" for Maven group and Java package

pnepywoda:
lets just take the backcompat break in BearerTokens.fromPaths but keep
the package the same for the super-widely-used AuthHeader and
BearerToken classes

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,8 @@ Provides a filter to inject user identifier information into slf4j and Jetty log
 Gradle:
 ```
 dependencies {
-    compile "com.palantir.tokens2:auth-tokens:<version>"
-    compile "com.palantir.tokens2:auth-tokens-filter:<version>"
+    compile "com.palantir.tokens:auth-tokens:<version>"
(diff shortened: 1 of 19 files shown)


markelliot:
Slightly stronger remediation: just delete the BearerTokens class from this repo altogether?

markelliot:
Minimally: publish it in a different project with a group name that includes the "2"

pnepywoda:
I would also be ok with deleting. Though is that better than just changing
the return value?

diogoholanda:
I would have been fine with the proposal here originally.
Defer to @markelliot on the decision here.
Keep in mind that some internal projects have already made the move

pnepywoda:
I only know of one internal project, but it's still worth changing them. It's like 1 or maybe 2 have switched where are 50 have not

pnepywoda:
I'll change whatever you guys want, I prefer keeping the method with returning java8 Optional, but we really should push this through. It's making migration to http-remoting3 really inconvenient

PR #51, opened 2017-10-04
Ignore unknown props on JWT deserialization

shashiadi:
Explicitly add the deserialization feature to the mapper in UnverifiedJWT - otherwise this blows up for polymorphic tokens which encode type information in a special field called "typ" (alternatively can just add this explicitly as an Optional<String>).


uschi2000 (approved):
Looks good to me. Should at some point make it more obvious in the test what those encoded JWT triples are so one can read and understand the tests.

PR #63, opened 2018-06-08
Constant time equals method

j-baker:
Makes timing attacks impossible if e.g. you have a `Map<AuthHeader, Some permissions granted>`

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
@@ -18,6 +18,8 @@ package com.palantir.tokens.auth;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
 import java.util.regex.Pattern;
 import org.immutables.value.Value;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -39,6 +41,12 @@
     @JsonValue
     public abstract String getToken();
 
+    @Value.Derived
+    @Value.Auxiliary
+    byte[] getTokenAsBytes() {
```
iamdanfox: Any reason this is package accessible or could this be made `protected`?

j-baker: protected is actually more accessible than package?

j-baker: protected = package but also subclasses can access

iamdanfox:
Just a thought - this is going to double the memory footprint of BearerTokens right?

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -39,6 +41,12 @@
     @JsonValue
     public abstract String getToken();
 
+    @Value.Derived
+    @Value.Auxiliary
```
iamdanfox: Given that we already have our own `equals` `hashCode` and `toString` - this `@Value.Auxiliary` doesn't actually do anything right? https://immutables.github.io/immutable.html#auxiliary-attributes

j-baker:
yes, but I think the extra 200 bytes ram is worth it. I could also do the conversion on equals, but this is an expensive conversion according to profiling i've done in the past. To be clear, right now it's a bit risky to have the cache I describe - on services with little traffic it's likely possible you can infer bearer tokens remotely right now.

iamdanfox (approved):

iamdanfox:
Looks like I don't have merge perms on this repo - maybe @uschi2000 can help?

j-baker:
Good news! I do.

PR #66, opened 2018-06-27
Check header look like JWT before processing

snegrus-adb:


snegrus-adb:
@j-baker @sandorw for review

sandorw (approved):
Looks good!

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFilter.java
```diff
@@ -69,6 +72,13 @@ private void clearMdc() {
         MDC.remove(TOKEN_ID_KEY);
     }
 
+    /**
+     * Based on the structure check from {@link UnverifiedJsonWebToken#of(BearerToken)}.
+     */
+    private boolean hasJwtStructure(String rawAuthHeader) {
+        return rawAuthHeader.split("\\.").length == 3;
```
j-baker: instead of copying the string and splitting into 3 every time, any chance we could just do:

    return rawAuthHeader.chars().filter(x -> x == '.').count() == 2;

?

j-baker: it might be worth adding a 'UnverifiedJsonWebToken.tryParse(String)' instead, btw - it avoids putting this logic into this class (which probably shouldn't care). It can do the same check and then delegate to 'of' and then return Optional.empty, which would let us avoid the catching Throwable here as well.

snegrus-adb: Wouldn't in that case `tryParse` be another version of `.of` method but without the exception?

snegrus-adb:
@j-baker is there anything else needed for this PR?

auth-tokens/src/test/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenTests.java
```diff
@@ -90,4 +98,10 @@ public void invalidJwt_invalidPayloadToken() {
             assertEquals("Invalid JWT: cannot parse payload", e.getMessage());
         }
     }
+
+    private void assertValidToken(UnverifiedJsonWebToken token) {
```
j-baker: these tests aren't just asserting that the token's valid - they're asserting facts about what's in the token. For example, the session token test is asserting that the token is valid, too.

This method should be called 'assertValidApiToken'

auth-tokens/src/test/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenTests.java
```diff
@@ -58,17 +58,25 @@ public void testAsJwt_validJwtFromSessionToken() {
...
+
+    @Test
+    public void testAsJwt_validJwtFromParsedToken() {
+        Optional<UnverifiedJsonWebToken> token = UnverifiedJsonWebToken.tryParse(PROXY_TOKEN.getToken());
+        assertValidToken(token.get());
+    }
+
+    @Test
+    public void invalidJwt_parseReturnsEmpty() {
+        Optional<UnverifiedJsonWebToken> parsedJwt = UnverifiedJsonWebToken.tryParse(INVALID_BEARER_TOKEN.getToken());
```
j-baker, line 78: worth a test for the case where the token has two periods in, but still isn't a valid bearer token

j-baker (approved):

PR #68, opened 2018-07-04
Much faster validation

j-baker:
To take a String header, convert it into a bearer token, and then turn it into
an unverified JWT takes 5400ns. Additionally the method itself requires the
bearer token, which adds another 3600ns - in total we spend 9us decoding
bearer tokens on each call. This is actually significant; if we want a service
to do 10k RPCs per core, they can expect roughly 9% of all CPU cycles to
be spent decoding bearer tokens.

This PR drops the overhead significantly; it hand codes the validation
of bearer tokens (which drops bearer token parsing to 450ns from 3600ns),
replaces UUID.toString (which is horrible) with a custom implementation
which saves about 450ns on each decode, and lastly replaces the base64
decoder with one from XML which runs 50ns faster (for free).

So, before: 5400ns, after: 1480ns

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
@@ -18,7 +18,6 @@ package com.palantir.tokens.auth;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
-import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.util.BitSet;
 import org.immutables.value.Value;
@@ -54,19 +53,17 @@ public abstract class BearerToken {
     @JsonValue
     public abstract String getToken();
 
-    /**
-     * isValidBearerToken validates that this is safe.
-     */
+    // We use a hand-written getBytes() implementation for performance reasons.
+    // Note that we don't need to worry about the character set (e.g., UTF-8) because
+    // the set of allowable characters are single bytes.
     @Value.Derived
     @SuppressWarnings("DesignForExtension")
     byte[] getTokenAsBytes() {
         String token = getToken();
         byte[] result = new byte[token.length()];
-
         for (int i = 0; i < result.length; i++) {
             result[i] = (byte) token.charAt(i);
         }
-
         return result;
     }
 
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
@@ -20,7 +20,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
-import java.util.regex.Pattern;
+import java.util.BitSet;
 import org.immutables.value.Value;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


j-baker:
Also, while 10us may sound like an eatable amount of error, internally I've noticed a lot of places which casually add 5-10 microseconds, and they can add up. If you factor in metrics + logging, even at 1k requests per second you're spending >20ms per second CPU time doing this work right now, and I'd like to reduce these overheads where they exist.

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -35,7 +35,20 @@
     private static final Logger log = LoggerFactory.getLogger(BearerToken.class);
 
     private static final String VALIDATION_PATTERN_STRING = "^[A-Za-z0-9\\-\\._~\\+/]+=*$";
```
uschi2000, line 36: would be good to add a few more edge cases now that we're hand jamming the regex:
```
a
abc
a===
```

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -51,13 +64,26 @@
...
         }
         return ImmutableBearerToken.of(token);
     }
 
+    private static boolean isValidBearerToken(String token) {
+        boolean isAtEnd = false;
+        int length = token.length();
+        for (int i = 0; i < length; i++) {
+            char character = token.charAt(i);
+            if ((isAtEnd && character != '=') || !allowedCharacters.get(character)) {
```
uschi2000: while I think this is correct, it took me a while to convince myself. i guess the tests should be sufficient to make this stable, but wondering if a more procedural algorithm (with nested if/else etc) would make this easier to follow?

auth-tokens/src/main/java/com/palantir/tokens/auth/UuidStringConverter.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Heavily based on Jackson's 'UUIDSerializer'. Changes made have been to change the format to comply
+ * with Palantir style, and make the method return a string rather than writing into a JSON processor.
+ *
+ * That project is Apache 2.0 licensed, but the original source contains no license header on its own,
+ * so here we link to their license file.
+ *
+ * https://github.com/FasterXML/jackson-databind/blob/master/src/main/resources/META-INF/LICENSE
+ */
+class UuidStringConverter {
```
uschi2000, line 30: want to add 1-2 sanity tests?

j-baker: sure

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -51,13 +64,26 @@
...
         AuthTokensPreconditions.checkArgument(!token.isEmpty(), "BearerToken cannot be empty");
-        if (!VALIDATION_PATTERN.matcher(token).matches()) {
+        if (!isValidBearerToken(token)) {
             log.trace("Error parsing BearerToken, must match pattern {}: {}", VALIDATION_PATTERN_STRING, token);
             throw new IllegalArgumentException("BearerToken must match pattern " + VALIDATION_PATTERN_STRING);
         }
         return ImmutableBearerToken.of(token);
     }
 
+    private static boolean isValidBearerToken(String token) {
```
uschi2000: oh, one more thing: can you add:
```
// Optimized implementation of the regular expression VALIDATION_PATTERN_STRING
```

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
j-baker: hopefully improved

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
j-baker: done

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
j-baker: done

uschi2000 (approved):

PR #69, opened 2018-07-13
Fix checkstyle and failing tests

ellisjoe:
Tests didn't run as part of https://github.com/palantir/auth-tokens/pull/68 so this fixes up the broken bits.

As for the base64 decoder: it appears the the DatatypeConverter implements `xsd:base64Binary` which is based on RFC 2045 while the java.util decoder implements RFC4648. Haven't looked into the differences but there are tests in `UnverifiedJsonWebTokenTests` that fail with the DatatypeConverter.


auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -80,18 +79,27 @@ public static BearerToken valueOf(String token) {
...
     private static boolean isValidBearerToken(String token) {
-        boolean isAtEnd = false;
         int length = token.length();
-        for (int i = 0; i < length; i++) {
-            char character = token.charAt(i);
-            if (isAtEnd && character != '=') {
-                return false;
-            } else if (!allowedCharacters.get(character)) {
-                return false;
+        int cursor;
```
j-baker: maybe explicitly initialize here?

j-baker: be more similar to the below one

auth-tokens/src/main/java/com/palantir/tokens/auth/UuidStringConverter.java
```diff
@@ -37,10 +37,9 @@ static String toString(UUID value) {
         final long msb = value.getMostSignificantBits();
         writeInt((int) (msb >> 32), ch, 0);
         ch[8] = '-';
-        int i = (int) msb;
-        writeShort(i >>> 16, ch, 9);
+        writeShort((int) msb >>> 16, ch, 9);
```
j-baker: please can you revert these changes? I find it difficult to read re: the precedence between (int) (msb >>> 16) and ((int) msb) >>> 16)

j-baker:
did tests run on this PR? Because if they didn't, we should probably work out what's happening

ellisjoe:
^ yeah you can see two checks pass with one of them being circle while on the other PR only one check ran

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
ellisjoe: done

auth-tokens/src/main/java/com/palantir/tokens/auth/UuidStringConverter.java
ellisjoe: casting to an int first or second doesn't matter in this case. I switched this to add explicit parenthesis. also switched the above `>>` to `>>>` since it's the only one of the four bit shifts that uses the signed shift (which again doesn't matter).

auth-tokens/src/main/java/com/palantir/tokens/auth/UuidStringConverter.java
```diff
@@ -27,20 +27,19 @@
...
 
     static String toString(UUID value) {
         final char[] ch = new char[36];
         final long msb = value.getMostSignificantBits();
-        writeInt((int) (msb >> 32), ch, 0);
+        writeInt((int) (msb >>> 32), ch, 0);
         ch[8] = '-';
-        int i = (int) msb;
-        writeShort(i >>> 16, ch, 9);
+        writeShort((int) (msb >>> 16), ch, 9);
```
j-baker: this is definitely not equivalent to what was there before

j-baker: explicitly, (int) (-2^32 >>> 16) = 0, but ((int) -2^32)  >>> 16 = 2^16

j-baker: please can we just leave this logic the way it was? this is directly copied from Jackson and given that I don't have a decent test suite I'm not convinced of safety

ellisjoe: so they're not equivalent ints but they are equivalent shorts, which is what we care about here. fine going back to the original logic and just giving the variable a real name though so we keep this as close to the original as possible.

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -80,18 +79,27 @@ public static BearerToken valueOf(String token) {
...
-            char character = token.charAt(i);
-            if (isAtEnd && character != '=') {
-                return false;
-            } else if (!allowedCharacters.get(character)) {
-                return false;
+        int cursor = 0;
+
+        for (; cursor < length; cursor++) {
+            if (!allowedCharacters.get(token.charAt(cursor))) {
+                break;
```
iamdanfox, line 87: This feels a bit subtle, that we re-use the cursor to validate a run of `===` at the end...

ellisjoe: suggestions? I found this far more readable than the previous flag that determined whether we were validating the run of `==` at the end or the initial run of valid characters

iamdanfox: I wonder if we started validating at the end and step backwards would it look simpler?

ellisjoe: I'd still end up using the same cursor for validating both parts though, no?

iamdanfox: yeah dw dude, if the tests pass then we're all good!

iamdanfox (approved):

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
ellisjoe: cool. I personally thought this read nicely so if you have an alternative that you think is easier to read and performant (which was the whole reason for this change) then happy to follow up

PR #71, opened 2018-07-25
Upgrade baseline

ellisjoe:


gradle.properties
```diff
@@ -1 +1 @@
-baselineVersion = 0.14.0
+baselineVersion = 0.15.1
```
iamdanfox, line 1: I'd recommend getting rid of this file entirely - the indirection here isn't really that useful!

iamdanfox (approved):

gradle.properties
ellisjoe: agreed

PR #80, opened 2018-12-11
Performance improvements parsing UnverifiedJsonWebToken

carterkozak:
Implemented JMH benchmarks for UnverifiedJsonWebToken.tryParse.

Results before this change:
```
UnverifiedJsonWebTokenBenchmarks.parseNonJwt        thrpt    3  8530177.987 ± 438717.090  ops/s
UnverifiedJsonWebTokenBenchmarks.parseSessionToken  thrpt    3   502692.902 ±  19562.944  ops/s
```

With this change:
```
UnverifiedJsonWebTokenBenchmarks.parseNonJwt        thrpt    3  14073771.303 ± 1743291.960  ops/s
UnverifiedJsonWebTokenBenchmarks.parseSessionToken  thrpt    3    662978.619 ±   66605.517  ops/s
```

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
@@ -20,6 +20,7 @@ package com.palantir.tokens.auth;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import java.io.IOException;
 import java.nio.ByteBuffer;
@@ -44,9 +45,10 @@ import org.slf4j.LoggerFactory;
 @ImmutablesStyle
 public abstract class UnverifiedJsonWebToken {
 
-    private static final ObjectMapper MAPPER = new ObjectMapper()
+    private static final ObjectReader READER = new ObjectMapper()
             .registerModule(new Jdk8Module())
-            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+            .readerFor(JwtPayload.class);
 
     private static final Logger log = LoggerFactory.getLogger(UnverifiedJsonWebToken.class);
 
@@ -75,7 +77,7 @@ public abstract class UnverifiedJsonWebToken {
      * before attempting to create an {@link UnverifiedJsonWebToken}.
      */
     public static Optional<UnverifiedJsonWebToken> tryParse(String rawAuthHeader) {
-        if (rawAuthHeader.chars().filter(x -> x == '.').count() == 2) {
+        if (countCharacter(rawAuthHeader, '.') == 2) {
             try {
                 return Optional.of(of(AuthHeader.valueOf(rawAuthHeader).getBearerToken()));
             } catch (Throwable t) {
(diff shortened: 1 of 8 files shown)


markelliot:
We probably ought to rerun this with AfterBurner enabled.

carterkozak:
I tested with afterburner, it didn't make much difference in the current implementation, however combined with a few other optimizations it can increase the `parseSessionToken` results from ~600k to ~900k ops/sec. I didn't push those because the implementation sacrifices readability. Happy to put those together if you're interested, probably works best in a separate PR.

carterkozak:
Sample implementation, needs to be cleaned up before it can be merged, but it's consistently providing >1m ops/sec
https://github.com/palantir/auth-tokens/commit/3510e81847c9a8c31b2a632752db5d31b1f7e5bc

auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
```diff
@@ -44,9 +45,10 @@
 @ImmutablesStyle
 public abstract class UnverifiedJsonWebToken {
 
-    private static final ObjectMapper MAPPER = new ObjectMapper()
+    private static final ObjectReader READER = new ObjectMapper()
             .registerModule(new Jdk8Module())
-            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```
ferozco, line 48: why do we ignore unknown properties?

benchmarks/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenBenchmarks.java
```diff
@@ -0,0 +1,58 @@
...
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+
+@Fork(1)
+@Threads(1)
+@Warmup(iterations = 10, time = 5)
+@Measurement(iterations = 3, time = 3)
+public class UnverifiedJsonWebTokenBenchmarks {
```
ferozco, line 33: Where do we plan on using these benchmarks?

ferozco (commented):
Just out of curiosity, what motivated this change?

benchmarks/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenBenchmarks.java
carterkozak: We don't have a great story for benchmark automation at the moment, so running benchmarks is a manual action.

auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
carterkozak: Allows us to add additional fields to the jwt payload in the future without breaking existing consumers.

benchmarks/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenBenchmarks.java
```diff
@@ -0,0 +1,58 @@
...
+        return UnverifiedJsonWebToken.tryParse(NOT_JWT);
+    }
+
+    @Benchmark
+    @BenchmarkMode(Mode.Throughput)
+    public final Optional<UnverifiedJsonWebToken> parseSessionToken() {
+        return UnverifiedJsonWebToken.tryParse(SESSION_TOKEN);
+    }
+
+    private static String repeat(CharSequence seq, int count) {
```
ellisjoe: can use `Strings.repeat(...)` from guava

ellisjoe (commented):
one minor nit

benchmarks/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenBenchmarks.java
carterkozak: We don't have guava in this module. I can add the dependency, but this was simpler.

ellisjoe (approved):

PR #81, opened 2018-12-11
Remove dependency on jackson-datatype-jdk8

carterkozak:
Only used for an internal class, we can use Optional.ofNullable
ourselves.

### DIFF
--- a/auth-tokens/build.gradle
+++ b/auth-tokens/build.gradle
@@ -2,7 +2,6 @@ apply from: "${rootDir}/gradle/publish.gradle"
 
 dependencies {
     compile "com.fasterxml.jackson.core:jackson-databind"
-    compile "com.fasterxml.jackson.datatype:jackson-datatype-jdk8"
     compile "org.slf4j:slf4j-api"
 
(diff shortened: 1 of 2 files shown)


dansanduleac:
What's the rationale for this change?
I imagine performance is similar since we still create an Optional only to throw it away right afterwards.

carterkozak:
Unnecessary jackson dependencies in widely consumed libraries make me uneasy because projects tend to pin specific dependency versions. We don't want mis-matched versions of jackson-databind and jackson-datatype-jdk8. At some point jackson-datatype-jdk8 will be removed like jackson-datatype-jdk7, which still causes problems in many libraries (atlas for example).

dansanduleac:
Cool, thanks for the explanation!
FWIW, I think the pinning aspect will not be a problem once we switch off of nebula.dependency-recommender and use gradle platforms instead, which make this sort of thing much more manageable.
That being said, the change is not very invasive so I guess it's fine to make this tradeoff.

ellisjoe (approved):

PR #85, opened 2019-01-16
Upgrade build tooling

robert3005:
Upgrades gradle and all of gradle plugins that we use in this project

cc @iamdanfox


carterkozak (approved):
Looks reasonable to me

auth-tokens/src/test/java/com/palantir/tokens/auth/BearerTokenTests.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ * (c) Copyright 2016 Palantir Technologies Inc. All rights reserved.
```
ellisjoe, line 2: odd that it didn't update the year

robert3005: I have a handy sed command that doesn't update the year since we shouldn't

iamdanfox (approved):

ferozco (approved):

PR #87, opened 2019-01-16
Configure mdc auth properties from cookie auth

robert3005:


auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFilter.java
```diff
@@ -41,17 +43,34 @@
     public static final String SESSION_ID_KEY = "sessionId";
     public static final String TOKEN_ID_KEY = "tokenId";
 
+    private final String cookieAuthKey;
+
+    public BearerTokenLoggingFilter(String cookieAuthKey) {
+        this.cookieAuthKey = cookieAuthKey;
```
carterkozak: this is a break, should be optional/default with a no-arg ctor.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFilter.java
```diff
@@ -41,17 +43,34 @@
     public static final String SESSION_ID_KEY = "sessionId";
     public static final String TOKEN_ID_KEY = "tokenId";
 
+    private final String cookieAuthKey;
```
carterkozak: Conjure allows endpoints to specify different cookie names, how do we plan to register a separate filter instance for each endpoint?

carterkozak (changes requested):

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFilter.java
robert3005: ok

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFilter.java
robert3005: I believe we fall on convention here - just like I could define authorization to be under different auth headers. Unless you want to get in the game of generating this logging handler for every endpoint separately but right now our server infra doesn't allow that in all cases and you have already handled the other case

carterkozak: >  just like I could define authorization to be under different auth headers

Conjure only allows the `Authorization` to be used for header auth.

> Unless you want to get in the game of generating this logging handler for every endpoint separately

I already am, our undertow handlers have already solved this problem  ;-) 

I think we can solve this for jersey using a DynamicFeature to check the first arg for an auth cookie, without requiring per-app customization or additional code generation.

robert3005: Sure, while I agree with you I don't think logging is covered under spec in any form. Once it added we can adjust it. You'll notice that it's only our server implementation uses this filter and not the conjure-java-runtime part. 

Don't get me wrong - I think what you're proposing is probably the right thing to do just don't think it's valuable right now

iamdanfox (approved):
I did a code-search internally and it seems that all `@CookieParam` usages internally use the same value, so perhaps it's OK to just hard-code that in WC?

carterkozak:
@iamdanfox Did your search include conjure definitions using auth:header as well?

iamdanfox:
Yeah - interestingly I actually didn't find usages of authType: "cookie" in conjure definitions (all the `authType: "header"` ones are covered by our existing BearerTokenLoggingFilter).

iamdanfox:
OK I've put up a quick PR to explore the idea of using a DynamicFeature: https://github.com/palantir/auth-tokens/pull/90.

robert3005:
gonna close in favour of #90

PR #90, opened 2019-01-17
Cookie logging

iamdanfox:
## Before this PR

Incoming requests that use cookies for auth don't get the MDC values or RequestContext populated for `userId`, `sessionId` or `tokenId`.

## After this PR

Users can register a single dynamic feature which will set up the right MDC values appropriately:

```diff
-resourceConfig.register(BearerTokenLoggingFilter.class)
+resourceConfig.register(BearerTokenLoggingFeature.class)
```

This is a jax-rs DynamicFeature, which checks each endpoint and wires up either header, cookie or no filter. This should handle even the trickiest hand-rolled jax-rs endpoints, including this one I found:

```
    @GET
    @Timed
    @Path("logout")
    public Response logout(
            @CookieParam(CookieManager.BROWSERID) BrowserId browserId,
            @CookieParam(CookieManager.TOKEN) BearerToken sessionToken,
            @Context HttpServletRequest request,
            @Context HttpServletResponse response) {
```

### DIFF
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenCookieLoggingFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenCookieLoggingFilter.java
@@ -28,11 +28,11 @@ import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
 @Priority(Priorities.AUTHORIZATION)
-class CookieBearerTokenLoggingFilter implements ContainerRequestFilter {
+class BearerTokenCookieLoggingFilter implements ContainerRequestFilter {
     private static final Logger log = LoggerFactory.getLogger(BearerTokenLoggingFilter.class);
     private final String cookie;
 
-    public CookieBearerTokenLoggingFilter(String cookie) {
+    BearerTokenCookieLoggingFilter(String cookie) {
         this.cookie = cookie;
     }
 
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenCookieLoggingFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenCookieLoggingFilter.java
@@ -1,5 +1,5 @@
 /*
- * (c) Copyright 2017 Palantir Technologies Inc. All rights reserved.
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFeature.java
```diff
@@ -0,0 +1,84 @@
...
+            return;
+        }
+
+        List<Parameter> cookieParams = Arrays.stream(parameters)
+                .filter(param -> param.getAnnotation(CookieParam.class) != null)
+                .filter(param -> param.getType().isAssignableFrom(BearerToken.class))
+                .collect(toList());
+
+        if (cookieParams.size() > 1) {
+            throw new IllegalStateException("Multiple BearerToken parameters annotated with @CookieParam");
```
carterkozak: Maybe `log.warn` and use the first one? Not sure we want to fail catastrophically here

carterkozak: (not a strong preference)

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenCookieLoggingFilter.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    private static void setUnverifiedContext(ContainerRequestContext requestContext, String key, String value) {
+        MDC.put(key, value);
+        requestContext.setProperty(getRequestPropertyKey(key), value);
+    }
+
+    public static String getRequestPropertyKey(String key) {
+        return "com.palantir.tokens.auth." + key;
+    }
```
carterkozak, line 50: maybe extract these static methods to a shared package-private utility class?

carterkozak (approved):
Commented some nits, but this looks good to merge as is. Up to you whether or not to address comments.

versions.props
```diff
@@ -1,13 +1,13 @@
-com.fasterxml.jackson.*:jackson-* = 2.6.7
+com.fasterxml.jackson.*:jackson-* = 2.8.9
```
carterkozak, line 1: Any reason not to use the latest 2.9.8?

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFeature.java
```diff
@@ -0,0 +1,84 @@
...
+                    resourceInfo.getResourceMethod());
+            context.register(new BearerTokenCookieLoggingFilter(cookieName));
+            return;
+        }
+
+        log.debug(
+                "Not adding BearerTokenLoggingFilter or BearerTokenCookieLoggingFilter as no "
+                        + "@HeaderParam or @CookieParam annotated arguments were found",
+                resourceInfo.getResourceClass(),
+                resourceInfo.getResourceMethod());
```
carterkozak: log statement is not parameterized, the logger may ignore parameter values. Do we have safe-logging on the classpath?

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFeature.java
```diff
@@ -0,0 +1,84 @@
...
+                .collect(toList());
+
+        if (authorizationHeaderParams.size() > 1) {
+            throw new IllegalStateException("Multiple parameters annotated with @HeaderParam('Authorization')");
+        }
+
+        if (authorizationHeaderParams.size() == 1) {
+            log.debug("Enabling BearerTokenLoggingFilter",
+                    resourceInfo.getResourceClass(),
+                    resourceInfo.getResourceMethod());
```
carterkozak: parameters may be dropped, needs `{}` and/or safe-logging `Arg`

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFeature.java
```diff
@@ -0,0 +1,84 @@
...
+
+        if (cookieParams.size() > 1) {
+            throw new IllegalStateException("Multiple BearerToken parameters annotated with @CookieParam");
+        }
+
+        if (cookieParams.size() == 1) {
+            String cookieName = cookieParams.get(0).getAnnotation(CookieParam.class).value();
+            log.debug("Enabling BearerTokenCookieLoggingFilter",
+                    resourceInfo.getResourceClass(),
+                    resourceInfo.getResourceMethod());
```
carterkozak: same

versions.props
iamdanfox: This one matches the dropwizard BOM - I didn't want to impose any unnecessarily high dependencies just because this is so low in the stack.

https://github.com/dropwizard/dropwizard/blob/v1.1.3/dropwizard-bom/pom.xml#L27

carterkozak: Sounds reasonable, thanks!

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFeature.java
iamdanfox: Have just added safe-logging to the filter, I figure everyone who uses this likely has it on their classpath already!

carterkozak (approved):
lgtm, thanks!

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFeature.java
```diff
@@ -0,0 +1,89 @@
...
+            log.debug("Enabling BearerTokenLoggingFilter {} {}",
+                    SafeArg.of("class", resourceInfo.getResourceClass()),
+                    SafeArg.of("method", resourceInfo.getResourceMethod()));
+            context.register(BearerTokenLoggingFilter.class);
+            return;
+        }
+
+        List<Parameter> cookieParams = Arrays.stream(parameters)
+                .filter(param -> param.getAnnotation(CookieParam.class) != null)
+                .filter(param -> param.getType().isAssignableFrom(BearerToken.class))
```
iamdanfox, line 66: this might be a bit overly-strict, as it won't catch any `@CookieParam("TOKEN") String` parameters...

carterkozak: I think this is fine for the `auth-tokens-filter`, if you're not using this library, I wouldn't expect this library to help you.

PR #93, opened 2019-01-28
Fix BearerTokenLoggingFeature annotation detection and loading

robert3005:
# Before this PR
You couldn't use BearerTokenLoggingFeature due to class loading cycles and annotation detection not working for interfaces
# After this PR
Both should work correctly

### DIFF
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFilter.java
@@ -37,6 +37,10 @@ import org.slf4j.MDC;
 public class BearerTokenLoggingFilter implements ContainerRequestFilter {
     private static final Logger log = LoggerFactory.getLogger(BearerTokenLoggingFilter.class);
 
+    public static final String USER_ID_KEY = Utilities.Key.USER_ID.getMdcKey();
+    public static final String SESSION_ID_KEY = Utilities.Key.SESSION_ID.getMdcKey();
+    public static final String TOKEN_ID_KEY = Utilities.Key.TOKEN_ID.getMdcKey();
+
     @Override
     public final void filter(ContainerRequestContext requestContext) {
         Utilities.clearMdc();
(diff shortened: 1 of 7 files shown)


robert3005:
@iamdanfox

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFilter.java
```diff
@@ -37,10 +37,6 @@
 public class BearerTokenLoggingFilter implements ContainerRequestFilter {
     private static final Logger log = LoggerFactory.getLogger(BearerTokenLoggingFilter.class);
 
-    public static final String USER_ID_KEY = Utilities.Key.USER_ID.getMdcKey();
-    public static final String SESSION_ID_KEY = Utilities.Key.SESSION_ID.getMdcKey();
-    public static final String TOKEN_ID_KEY = Utilities.Key.TOKEN_ID.getMdcKey();
```
iamdanfox: Given that these were public static final, could we just leave them as they used to be and annotate them @deprecated? We don't want to accidentally break someone!

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFilter.java
```diff
@@ -54,8 +50,4 @@ public final void filter(ContainerRequestContext requestContext) {
         Optional<UnverifiedJsonWebToken> parsedJwt = UnverifiedJsonWebToken.tryParse(rawAuthHeader);
         Utilities.recordUnverifiedJwt(requestContext, parsedJwt);
     }
-
-    public static String getRequestPropertyKey(String key) {
```
iamdanfox: Same here- would prefer not to delete if possible

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/Utilities.java
```diff
@@ -52,6 +52,10 @@ private static void setUnverifiedContext(ContainerRequestContext requestContext,
         }
     }
 
+    public static String getRequestPropertyKey(String key) {
```
iamdanfox: package private?

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFilter.java
iamdanfox: A code search internally found people relying on this:

```
Object userId = requestContext.getProperty(
BearerTokenLoggingFilter.getRequestPropertyKey(BearerTokenLoggingFilter.USER_ID_KEY));
```

robert3005: ok, we can keep just have to duplicate

iamdanfox (approved):

PR #143, opened 2019-07-09
Convert to gradle-consistent-versions

dansanduleac:
## Before this PR

Using the verbose and fiddly `nebula.dependency-lock`.

## After this PR
==COMMIT_MSG==
Lock dependencies using gradle-consistent-versions.
==COMMIT_MSG==


ferozco:
👍

PR #145, opened 2019-07-12
Convert to safe-loggable logs and preconditions

dansanduleac:
## Before this PR

A bunch of exceptions and logs did not use safe-loggable constructs, and so would be redacted.

## After this PR
==COMMIT_MSG==
Use safe-loggable log arguments and preconditions everywhere.
==COMMIT_MSG==


auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
```diff
@@ -110,13 +108,13 @@ private static int countCharacter(String input, char toCount) {
...
      */
     public static UnverifiedJsonWebToken of(BearerToken token) {
-        List<String> segments = Splitter.on('.').splitToList(token.getToken());
+        String[] segments = token.getToken().split("\\.");
         Preconditions.checkArgument(
-                segments.size() == 3,
+                segments.length == 3,
                 "Invalid JWT: expected 3 segments",
-                SafeArg.of("segmentsCount", segments.size()));
+                SafeArg.of("segmentsCount", segments.length));
```
carterkozak: this is a hot path, worth inverting this into `if (segments !=3) { /*work + throw */}`?

carterkozak (approved):

auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
dansanduleac: won't the JVM figure this out eventually?

carterkozak: Not likely, especially if we ever encounter a bad input.

dansanduleac: alright, changed

PR #146, opened 2019-07-15
Fix logging in BearerTokenLoggingFeature

dansanduleac:
## Before this PR

Logging `ResourceInfo.getResourceMethod()` directly, which throws an exception during serialization, concealing the error.

## After this PR
==COMMIT_MSG==
Fix logging in BearerTokenLoggingFeature.

`ResourceInfo.getResourceMethod()` is now turned into a string before being logged.
==COMMIT_MSG==

### DIFF
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFeature.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFeature.java
@@ -23,6 +23,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


carterkozak:
This has come up a lot in a large internal project, thoughts on a utility function in safe-logging to make this safer and easier? Ideally executing toString on the background thread.

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFeature.java
```diff
@@ -67,14 +68,14 @@ public void configure(ResourceInfo resourceInfo, FeatureContext context) {
...
-                    SafeArg.of("method", resourceInfo.getResourceMethod()));
+                    SafeArg.of("method", Objects.toString(resourceInfo.getResourceMethod())));
         }
 
         if (authorizationHeaderParams.isPresent() && authorizationHeaderParams.get().size() == 1) {
             log.debug(
                     "Enabling BearerTokenLoggingFilter",
                     SafeArg.of("class", resourceInfo.getResourceClass()),
-                    SafeArg.of("method", resourceInfo.getResourceMethod()));
+                    SafeArg.of("method", Objects.toString(resourceInfo.getResourceMethod())));
```
carterkozak, line 78: toString is more expensive than anything else here and can be avoided unless debug is enabled

dansanduleac: This only happens once, when you register a resource, so I'd actually be in favor of lowering this to info.

carterkozak: 👍

carterkozak: Would be nice to have a more general pattern than SafeArg.of("name", Objects.toString(value)), we've repeated that a lot.

carterkozak: Happy to decouple that from this PR though.

ferozco:
👍

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BearerTokenLoggingFeature.java
dansanduleac: Think more work has to be done in automating doing the right thing for sure, but I think it'll probably be easier to have different logic for different arg types (in the form of a chain of runtime type matchers) so that service or library authors don't have to think about whether their arg should be toString'ed or not.

PR #196, opened 2019-09-16
Add message arg to exceptions

pkoenig10:
## Before this PR
When an auth header is incorrect, the serialized error does not contain enough information for a client to know what was wrong. For example, providing a bearer token that does not match the required pattern results in the following response:
```
{
    "errorCode": "INVALID_ARGUMENT",
    "errorInstanceId": "5e876c95-ce13-4336-8bb0-4ac5716b1eea",
    "errorName": "Default:InvalidArgument",
    "parameters": {}
} 
```

For more context, see PDS-99495.

## After this PR
4xx errors include a `message` arg that provides information about why the request was invalid.

## Possible downsides?
I'm not convinced that duplicating the exception message as an arg is the correct solution, but providing this information to clients seems valuable. An alternative would be to include the exception message in the `SerializableError`, but it seems that we've [explicitly moved away](https://github.com/palantir/conjure-java-runtime-api/blob/2.3.0/errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java#L96) from doing that.

### DIFF
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
@@ -16,7 +16,9 @@
 
 package com.palantir.tokens.auth.http;
 
-import com.palantir.logsafe.Preconditions;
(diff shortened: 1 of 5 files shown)


pkoenig10:
Closing because these error messages do not provide enough additional information to warrant this change.

PR #228, opened 2019-11-08
Throw typed exception for malformed tokens

rshkv:
## Before this PR
Malformed tokens are thrown as (Safe)IllegalArgumentException which is too generic to handle. The reason for handling malformed tokens is e.g. to send a 401 status code to comply with the [OAuth spec](https://tools.ietf.org/html/rfc6750#section-3.1). Conjure e.g. will convert IllegalArgumentExceptions to 400 which isn't quite right in this case.

## After this PR
==COMMIT_MSG==
Throw `MalformedTokenException` for malformed bearer tokens.
==COMMIT_MSG==

cc @pkoenig10

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
@@ -23,6 +23,7 @@ import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.security.MessageDigest;
 import java.util.BitSet;
+import javax.ws.rs.NotAuthorizedException;
 import org.immutables.value.Value;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -74,8 +75,7 @@ public abstract class BearerToken {
         Preconditions.checkArgument(token != null, "BearerToken cannot be null");
         Preconditions.checkArgument(!token.isEmpty(), "BearerToken cannot be empty");
         if (!isValidBearerToken(token)) {
-            throw new SafeIllegalArgumentException("BearerToken must match pattern",
-                    SafeArg.of("validationPattern", VALIDATION_PATTERN_STRING));
+            throw new NotAuthorizedException("Bearer token must match pattern: " + VALIDATION_PATTERN_STRING);
         }
         return ImmutableBearerToken.of(token);
     }
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
@@ -71,11 +71,9 @@ public abstract class BearerToken {
     public static BearerToken valueOf(String token) {
         if (token == null) {
             throw new MalformedTokenException("Bearer token cannot be null");
-        }
-        if (token.isEmpty()) {
+        } else if (token.isEmpty()) {
             throw new MalformedTokenException("Bearer token cannot be empty");
-        }
-        if (!isValidBearerToken(token)) {
+        } else if (!isValidBearerToken(token)) {
             throw new MalformedTokenException(
                     "Bearer token must match pattern",
                     SafeArg.of("validationPattern", VALIDATION_PATTERN_STRING));
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
@@ -70,11 +70,11 @@ public abstract class BearerToken {
     @JsonCreator
     public static BearerToken valueOf(String token) {
         if (token == null) {
-            throw new MalformedTokenException("Bearer token cannot be null");
+            throw new MalformedAuthTokenException("Bearer token cannot be null");
         } else if (token.isEmpty()) {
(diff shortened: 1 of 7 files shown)


auth-tokens/src/main/java/com/palantir/tokens/auth/MalformedTokenException.java
```diff
@@ -0,0 +1,46 @@
...
+import com.palantir.logsafe.exceptions.SafeExceptions;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public final class MalformedTokenException extends IllegalArgumentException implements SafeLoggable {
+    private final String logMessage;
+    private final List<Arg<?>> arguments;
+
+    public MalformedTokenException(@CompileTimeConstant String message, Arg<?>... arguments) {
```
carterkozak: Should this be package private? Is there any reason to allow `MalformedTokenException` to be created outside of this library?

auth-tokens/src/main/java/com/palantir/tokens/auth/MalformedTokenException.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.google.errorprone.annotations.CompileTimeConstant;
+import com.palantir.logsafe.Arg;
+import com.palantir.logsafe.SafeLoggable;
+import com.palantir.logsafe.exceptions.SafeExceptions;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+public final class MalformedTokenException extends IllegalArgumentException implements SafeLoggable {
```
carterkozak: Mind adding class level javadoc describing this type? Given this library is called `auth-tokens` perhaps we should call this `MalformedAuthTokenException`?

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -71,10 +69,13 @@
...
-        Preconditions.checkArgument(!token.isEmpty(), "BearerToken cannot be empty");
-        if (!isValidBearerToken(token)) {
-            throw new SafeIllegalArgumentException("BearerToken must match pattern",
+        if (token == null) {
+            throw new MalformedTokenException("Bearer token cannot be null");
+        } else if (token.isEmpty()) {
+            throw new MalformedTokenException("Bearer token cannot be empty");
+        } else if (!isValidBearerToken(token)) {
+            throw new MalformedTokenException(
+                    "Bearer token must match pattern",
```
carterkozak: May want to avoid modifying these messages in this change to minimize potential failures from message equality checks (we've had a few of those lately).

carterkozak (commented):
In general I like the idea of having more specific exceptions, but I'm not sure this is the right way to solve the OAuth2 problem.
If we map MalformedTokenException to 401, we risk responding 401 instead of 400 when object `bearertoken` fields fail to deserialize properly, or when internal code fails to parse a service token.

We can update conjure code which parses the credential component to throw a 401 on failure instead of SafeIllegalArgumentException, if that would be helpful. Thoughts?

rshkv:
That sounds good. I understand in Conjure we'd still rather catch something specific than IllegalArgumentException.

pkoenig10 (changes requested):
Changing the exception type here is a breaking change and is not appropriate in all contexts where `BearerToken` is used.

Malformed tokens should be very rare in practice, so I'm not convinced it worth putting too much effort into improving the handling of that case.

If we do want to change the behavior, we should move this logic to the authorization extractor in Conjure. We won't get the behavior for Jersey resources but I don't think there's much we can do about that.

auth-tokens/src/main/java/com/palantir/tokens/auth/MalformedAuthTokenException.java
```diff
@@ -0,0 +1,49 @@
...
+import com.palantir.logsafe.SafeLoggable;
+import com.palantir.logsafe.exceptions.SafeExceptions;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Exception thrown by {@link BearerToken} for malformed tokens.
+ */
+public final class MalformedAuthTokenException extends IllegalArgumentException implements SafeLoggable {
```
markelliot, line 30: What if someone was previously catching SafeIllegalArgumentException?

pnepywoda: Are you suggesting that since this is a semantic break we bump the major version when releasing with this PR?

carterkozak:
I don’t think this change is required to resolve the original issue, there’s a separate question as to whether or not we should create a more specific exception for bad token constructors. I think the more specific exception is helpful, but I’m not confident it’s worth the risk of breaking consumers.

-ck

iamdanfox:
So concretely, I think we're suggesting a PR here would be preferable?

https://github.com/palantir/conjure-java/blob/aefca4e3e592f02bd094c91da6e01c12afab4230/conjure-java-undertow-runtime/src/main/java/com/palantir/conjure/java/undertow/runtime/ConjureAuthorizationExtractor.java#L52-L59

carterkozak:
Yep, and a similar change for jersey for mp.

-ck

pkoenig10:
Closing in favor of https://github.com/palantir/conjure-java/pull/656 and https://github.com/palantir/conjure-java-runtime/pull/1374.

PR #389, opened 2020-06-26
Fix: Replace javax dependencies with equivalent jakarta ones

robert3005:
## Before this PR
We use javax dependencies that are no longer developed and are replaced by jakarta equivalents

## After this PR
==COMMIT_MSG==
Replace javax dependencies with equivalent jakarta ones
==COMMIT_MSG==

Also bumped dropwizard testing to get rid of as many javax occurrences in tests as possible. The last remaining occurrence comes from jetty (via dropwizard testing)

## Possible downsides?
N/A


auth-tokens-filter/build.gradle
```diff
@@ -3,14 +3,15 @@ apply plugin: 'com.palantir.revapi'
...
-    compile "javax.ws.rs:javax.ws.rs-api"
+    compile "jakarta.servlet:jakarta.servlet-api"
+    compile "jakarta.annotation:jakarta.annotation-api"
+    compile "jakarta.ws.rs:jakarta.ws.rs-api"
     compile "org.slf4j:slf4j-api"
     implementation "com.palantir.safe-logging:preconditions"
 
     testCompile "io.dropwizard:dropwizard-testing"
     testCompile "junit:junit"
+    testCompile "org.assertj:assertj-core"
```
robert3005, line 14: this was previously implicitly coming from dropwizard-testing

carterkozak (approved):

PR #474, opened 2021-03-04
Use SOURCE retention for immutables style annotations

pkoenig10:
See https://github.com/immutables/immutables/issues/291.

Example of a similar fix in https://github.com/palantir/conjure-java-runtime/pull/1897.

This eliminates compilation warnings for consumers that do not directly depend on immutables.

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutablesStyle.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutablesStyle.java
@@ -23,6 +23,6 @@ import java.lang.annotation.Target;
 import org.immutables.value.Value;
 
 @Target({ElementType.PACKAGE, ElementType.TYPE})
-@Retention(RetentionPolicy.CLASS)
+@Retention(RetentionPolicy.SOURCE)
 @Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE, jdkOnly = true)
-public @interface ImmutablesStyle {}
+@interface ImmutablesStyle {}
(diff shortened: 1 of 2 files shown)


auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutablesStyle.java
```diff
@@ -23,6 +23,6 @@
 import org.immutables.value.Value;
 
 @Target({ElementType.PACKAGE, ElementType.TYPE})
-@Retention(RetentionPolicy.CLASS)
+@Retention(RetentionPolicy.SOURCE)
 @Value.Style(visibility = Value.Style.ImplementationVisibility.PACKAGE, jdkOnly = true)
-public @interface ImmutablesStyle {}
+@interface ImmutablesStyle {}
```
pkoenig10, line 28: Unfortunately, this annotation is used in a handful of internal repos.

Excavating out the change is a bit cumbersome because it may involve adding a dependency on our internal `immutables-style` package.

pnepywoda: so why change the visibility now? how about just fix the compile warning?

pkoenig10: Sure I guess I could open a second PR. I'd like to make both changes though.

pnepywoda: eh, if we're eventually making both no point in 2 prs

pnepywoda (approved):

PR #487, opened 2021-03-30
Use Base64 URL decoder

pkoenig10:
## Before this PR
`UnverifiedJsonWebToken` decodes token payloads using the standard Base64 decoding.

However, JWT payloads are encoding using the Base64 URL encoding.

In practice, this hasn't an issue because all of our claims are either numbers or Base64 encoded UUIDs. None of these characters will result in `-` or `/` in the token payload.

## After this PR
`UnverifiedJsonWebToken` uses the correct Base64 URL decoder to decode token payloads.

### DIFF
--- a/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
+++ b/auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
@@ -47,7 +47,6 @@ public class BasicAuthToBearerTokenFilter implements Filter {
     private static final Logger log = LoggerFactory.getLogger(BasicAuthToBearerTokenFilter.class);
 
     private static final String BASIC_AUTH_STR = "Basic";
-    private static final Base64.Decoder BASE_64_ENCODING = Base64.getUrlDecoder();
 
     @Override
     public void init(FilterConfig _filterConfig) {}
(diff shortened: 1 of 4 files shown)


pnepywoda (approved):

changelog/@unreleased/pr-487.v2.yml
```diff
@@ -0,0 +1,9 @@
+type: fix
+fix:
+  description: |-
+    Use Base64 URL decoder
+
+    ## Before this PR
```
carterkozak: Did you mean to commit this to the changelog?

pkoenig10: Nope, fixed

auth-tokens-filter/src/main/java/com/palantir/tokens/auth/http/BasicAuthToBearerTokenFilter.java
```diff
@@ -115,7 +114,7 @@ private AuthHeader base64DecodePassword(String rawAuthHeader) {
                 rawAuthHeader.substring(BASIC_AUTH_STR.length()).trim();
         String credentials;
         try {
-            credentials = new String(BASE_64_ENCODING.decode(base64Credentials), StandardCharsets.UTF_8);
```
carterkozak, line 118: static field was unnecessary because the getUrlDecoder method returns a static field -- indirection of a method invocation is insignificant.

carterkozak (approved):
lgtm, thanks

PR #753, opened 2022-08-01
Use case-insensitive matching for authentication scheme

pkoenig10:
## Before this PR
The `Authoziation` header authentication scheme is matching is case-sensitive.

But the authentication scheme is a case-insensitive value.

https://datatracker.ietf.org/doc/html/rfc7235#section-2.1

## After this PR
The `Authoziation` header authentication scheme is matching is case-insensitive.

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
@@ -36,8 +36,8 @@ public abstract class AuthHeader {
      * Takes the string form: "Bearer [token]" and creates a new {@link AuthHeader}.
      */
     public static AuthHeader valueOf(String authHeader) {
-        BearerToken bearerToken =
-                BearerToken.valueOf(authHeader.startsWith("Bearer ") ? authHeader.substring(7) : authHeader);
+        BearerToken bearerToken = BearerToken.valueOf(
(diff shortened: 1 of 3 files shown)


auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
```diff
@@ -36,8 +36,8 @@ public abstract class AuthHeader {
      * Takes the string form: "Bearer [token]" and creates a new {@link AuthHeader}.
      */
     public static AuthHeader valueOf(String authHeader) {
-        BearerToken bearerToken =
-                BearerToken.valueOf(authHeader.startsWith("Bearer ") ? authHeader.substring(7) : authHeader);
+        BearerToken bearerToken = BearerToken.valueOf(
+                authHeader.regionMatches(true, 0, "Bearer ", 0, 7) ? authHeader.substring(7) : authHeader);
```
carterkozak: Would be nice to describe the case-insensitive match and link rfc 7235 here as well, that way future-us have this context.
```suggestion
                authHeader.regionMatches(/* ignoreCase */ true, 0, "Bearer ", 0, 7) ? authHeader.substring(7) : authHeader);
```

pkoenig10: I generally prefer to rely on the Git history for these types of things, but happy to add comments in the code if you'd prefer.

carterkozak (approved):

PR #806, opened 2023-01-27
Add support for org claim

pkoenig10:
Add support for the `org` claim we are adding to our JWTs. This claim is also a Base64-encoded UUID.

For more context, see https://g.p.b/foundry/mp/pull/16056.

## Possible downsides?

This may break users of this library whose JWTs have an `org` claim that is not a Base64-encoded UUID. But we've never claimed that this library should be used for arbitrary JWTs - it's only intended to be used for Palantir JWTs.

### DIFF
--- a/auth-tokens/src/test/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenTests.java
+++ b/auth-tokens/src/test/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenTests.java
@@ -26,49 +26,50 @@ import org.junit.jupiter.api.Test;
 
 final class UnverifiedJsonWebTokenTests {
 
-    private static final BearerToken SESSION_TOKEN = BearerToken.valueOf("header."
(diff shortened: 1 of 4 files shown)


pkoenig10:
Do not merge until we've made the internal change to begin producing this claim.

auth-tokens/src/test/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenTests.java
```diff
@@ -26,49 +26,50 @@
 
 final class UnverifiedJsonWebTokenTests {
 
-    private static final BearerToken SESSION_TOKEN = BearerToken.valueOf("header."
```
pkoenig10, line 29: These tests are currently designed based on our internal token types (session vs API). This is an implementation details that consumers should not be concerned with. I've refactored these tests to better reflect the guarantees and assumptions of this library itself.

carterkozak (approved):
Thanks, I appreciate the updated documentation as well

carterkozak:
Feel free to merge this when you're ready

PR #1117, opened 2024-12-04
Fix build

schlosna:
## Before this PR

https://github.com/palantir/auth-tokens/pull/1116 excavator failed and broke the build

## After this PR

==COMMIT_MSG==
Fix build
==COMMIT_MSG==


carterkozak (approved):
Thanks! I've updated the branch protection rules for this repo to require `circle-all` and `changelog` as well, so this cannot happen again.
I'm not sure how this repo fell off the standard path

PR #1284, opened 2025-09-04
Add gradle jmh plugin

tdeitch:
## Before this PR

Running jmh benchmarks was [a manual process](https://github.com/palantir/auth-tokens/pull/80#discussion_r241114755).

## After this PR

Benchmarks can be run with `./gradlew jmh`.

==COMMIT_MSG==
Add gradle jmh plugin
==COMMIT_MSG==

## Possible downsides?

This still doesn't constitute benchmark automation, which would require integration with our test framework. It just makes it easier to manually run the benchmarks.


benchmarks/build.gradle
```diff
@@ -1,6 +1,8 @@
+apply plugin: 'me.champeau.jmh'
```
pkoenig10, line 1: This plugin has a number of problems. It is trivial to set this up ourselves - I can provide some examples where we've done this internally.

pkoenig10 (changes requested):

PR #1285, opened 2025-09-04
Parse and expose "exp" claim

tdeitch:
## Before this PR

We do not support the "exp" claim defined in [RFC 7519 § 4.1.4](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.4).

## After this PR

==COMMIT_MSG==
Parse and expose "exp" claim
==COMMIT_MSG==

## Possible downsides?

Some users might prefer the raw number value rather than the parsed NumericDate value.

## Benchmark

```
Before

Benchmark                                            Mode  Cnt         Score         Error  Units
UnverifiedJsonWebTokenBenchmarks.parseNonJwt        thrpt    3  28155737.632 ± 6352493.701  ops/s
UnverifiedJsonWebTokenBenchmarks.parseSessionToken  thrpt    3   1545347.909 ±  282668.819  ops/s

After

Benchmark                                            Mode  Cnt         Score         Error  Units
UnverifiedJsonWebTokenBenchmarks.parseNonJwt        thrpt    3  27780590.209 ± 4099857.530  ops/s
UnverifiedJsonWebTokenBenchmarks.parseSessionToken  thrpt    3   1460278.320 ±  108439.168  ops/s
```

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
@@ -26,8 +26,6 @@ import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
-import org.immutables.value.Value;
-
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.time.Instant;
@@ -36,6 +34,7 @@ import java.time.ZoneOffset;
 import java.util.Base64;
 import java.util.Optional;
 import java.util.UUID;
+import org.immutables.value.Value;
 
 /**
  * Represents the parsed form of a JWT but does not verify the token signature.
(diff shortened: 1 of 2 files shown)


auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
```diff
@@ -85,6 +88,14 @@ public abstract class UnverifiedJsonWebToken {
     @Value.Parameter
     public abstract Optional<String> getUnverifiedOrganizationId();
 
+    /**
+     * Returns the unverified expiration time, i.e. the "exp" claim, of the JWT
+     * or absent if the JWT does not contain the "exp" claim.
+     */
+    @Safe
+    @Value.Parameter
+    public abstract Optional<OffsetDateTime> getUnverifiedExpirationTime();
```
pkoenig10, line 97: `OffsetDateTime` is not the correct type to use here. This is just a offset-less timestamp so it should be `Instant`.

```suggestion
    public abstract Optional<Instant> getUnverifiedExpirationTime();
```

pkoenig10, line 97: I'm very wary about exposing this, for a couple reasons.
- The `exp` claim suggests to consumers that the token will be valid until that time. But this is not the case because tokens can be explicitly revoked before then. This is very common with short-lived API tokens generated for compute jobs, etc. This API is extremely easy to misuse in a way that will result in consumed accepting revoked tokens as valid.
- Not all tokens have `exp` claims, so consumers are still going to have to deal with the case when this claim is absent.

pkoenig10 (changes requested):

PR #1286, opened 2025-09-04
Support running benchmarks from Gradle and IDEs

pkoenig10:
Alternative to https://github.com/palantir/auth-tokens/pull/1284

This PR also update our usage of JMH to use the annotation processor instead of bytecode generation.

### DIFF
--- a/benchmarks/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenBenchmark.java
+++ b/benchmarks/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenBenchmark.java
@@ -17,19 +17,22 @@
 package com.palantir.tokens.auth;
 
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.BenchmarkMode;
 import org.openjdk.jmh.annotations.Fork;
-import org.openjdk.jmh.annotations.Measurement;
 import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
 import org.openjdk.jmh.annotations.Threads;
-import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.profile.GCProfiler;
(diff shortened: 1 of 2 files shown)


benchmarks/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenBenchmark.java
```diff
@@ -17,19 +17,22 @@
...
-import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.profile.GCProfiler;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
 
 @Fork(1)
 @Threads(1)
-@Warmup(iterations = 10, time = 5)
-@Measurement(iterations = 3, time = 3)
```
pkoenig10, line 31: JMH does 5 warmup iterations and 5 measurements by default, which is fine here.

benchmarks/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenBenchmark.java
```diff
@@ -38,13 +41,23 @@ public class UnverifiedJsonWebTokenBenchmarks {
 
     @Benchmark
     @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.SECONDS)
```
pkoenig10: I find it useful to be explicit about the output time unit.

aldexis: Thoughts on reducing the time-unit here? This outputs numbers in the millions of ops/s, which isn't super readable. It might be a bit nicer to output in ops/ms (or ms/op)

pkoenig10: Sounds good. I just did this to preserve the existing behavior, but I agree seconds probably isn't the best unit to be using here.

aldexis (approved):

PR #1331, opened 2025-10-24
Add support for svc claim

pkoenig10:
Add support for the `svc` claim we are adding to our JWTs.

For more context, see https://g.p.b/foundry/mp/pull/24561.

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
@@ -23,7 +23,7 @@ import org.immutables.value.Value;
  * Represents a HTTP authentication header. This class wraps a string in the form of "Bearer [token]".
  */
 @DoNotLog
-@Value.Immutable
+@Value.Immutable(builder = false, copy = false)
 @ImmutablesStyle
 // NOTE: no @JsonSerialize/@JsonDeserialize because auth headers are for use in @HeaderParam
 // see: https://jersey.java.net/apidocs/latest/jersey/javax/ws/rs/HeaderParam.html
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
@@ -20,14 +20,12 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
-import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.Safe;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import com.palantir.logsafe.logger.SafeLogger;
 import com.palantir.logsafe.logger.SafeLoggerFactory;
 import java.io.IOException;
-import java.nio.ByteBuffer;
 import java.util.Base64;
 import java.util.Optional;
 import java.util.UUID;
@@ -44,7 +42,7 @@ import org.immutables.value.Value;
  * logging.
  */
 @Safe
-@Value.Immutable
+@Value.Immutable(builder = false, copy = false)
 @ImmutablesStyle
 public abstract class UnverifiedJsonWebToken {
 
@@ -84,6 +82,14 @@ public abstract class UnverifiedJsonWebToken {
     @Safe
     @Value.Parameter
     public abstract Optional<String> getUnverifiedOrganizationId();
+
+    /**
+     * Returns the unverified first-party service name, i.e. the "svc" claim, of the JWT
+     * or absent if the JWT does not contain the "svc" claim.
+     */
+    @Safe
+    @Value.Parameter
+    public abstract Optional<String> getUnverifiedService();
 
     /**
      * Does a lower cost check on the structure of string provided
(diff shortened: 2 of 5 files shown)


auth-tokens/src/main/java/com/palantir/tokens/auth/AuthHeader.java
```diff
@@ -23,7 +23,7 @@
  * Represents a HTTP authentication header. This class wraps a string in the form of "Bearer [token]".
  */
 @DoNotLog
-@Value.Immutable
+@Value.Immutable(builder = false, copy = false)
```
pkoenig10, line 26: Drive-by to eliminate some generated code we do not use.

auth-tokens/src/main/java/com/palantir/tokens/auth/UnverifiedJsonWebToken.java
```diff
@@ -145,34 +152,21 @@ private static JwtPayload extractPayload(String payload) {
...
-    private static String decodeUuidBytes(byte[] bytes) {
-        Preconditions.checkArgument(
-                bytes.length == 16,
-                "Invalid JWT: cannot decode UUID, require 16 bytes",
-                SafeArg.of("bytesLength", bytes.length));
-        ByteBuffer byteBuffer = ByteBuffer.wrap(bytes);
-        long high = byteBuffer.getLong();
-        long low = byteBuffer.getLong();
-        return UuidStringConverter.toString(new UUID(high, low));
-    }
```
pkoenig10, line 162: Jackson's UUID deserializer support deserializing Base64-encoded UUIDs, so we don't need this. This avoids some additional allocations during deserailizing.

auth-tokens/src/test/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenTests.java
```diff
@@ -26,83 +26,101 @@
...
+    private static final BearerToken ALL_CLAIMS_TOKEN = BearerToken.valueOf("""
+        header.\
+        eyJzdWIiOiJ3NVAyV1FNQlEwNnB5WEl3U2xCLy9BPT0iLCJzaWQiOiJQOFpqMUQ1SVRlMjZUdGVLK1l1RFl3PT0iLCJqdGkiOiJwRm0wb1ZDSlQ\
+        rQ0dWZFhmMmJLMy9RPT0iLCJvcmciOiJGQlMycTgvbFQvMnNBRktxZ09pUW13PT0iLCJzdmMiOiJzZXJ2aWNlIn0\
+        .signature\
+        """);
+
+    private static final BearerToken REQUIRED_CLAIMS_TOKEN = BearerToken.valueOf("""
+        header.\
+        eyJzdWIiOiJ3NVAyV1FNQlEwNnB5WEl3U2xCLy9BPT0ifQ\
```
aldexis, line 38: Note to self: token went from `{"sub":"w5P2WQMBQ06pyXIwSlB//A==","exp": 1577865600}` to `{"sub":"w5P2WQMBQ06pyXIwSlB//A=="}`, which makes sense.

auth-tokens/src/test/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenTests.java
```diff
@@ -26,83 +26,101 @@
...
-            + "eyJzdWIiOiJrazlVMHB0ZVJ3K1FYYk55ZkZkcklBPT0iLCJqdGkiOiJ2MEtCNWdVTFJkT3dFWWh4Z1o3bERnPT0ifQo+"
-            + ".signature");
-
-    private static final BearerToken INVALID_PAYLOAD_TOKEN = BearerToken.valueOf("header."
-            + "eyJzdWIiOiJrazlVMHB0ZVJ3K1FYYk55ZkZkcklBPT0iLCJqdGkiOiJ2MEtCNWdVTFJkT3dFWWh4Z1o3bERnPT0iCg"
-            + ".signature");
+    private static final BearerToken ALL_CLAIMS_TOKEN = BearerToken.valueOf("""
+        header.\
+        eyJzdWIiOiJ3NVAyV1FNQlEwNnB5WEl3U2xCLy9BPT0iLCJzaWQiOiJQOFpqMUQ1SVRlMjZUdGVLK1l1RFl3PT0iLCJqdGkiOiJwRm0wb1ZDSlQ\
+        rQ0dWZFhmMmJLMy9RPT0iLCJvcmciOiJGQlMycTgvbFQvMnNBRktxZ09pUW13PT0iLCJzdmMiOiJzZXJ2aWNlIn0\
```
aldexis, line 32: (not necessarily for this PR - more of a meta-comment) Do you think it would be reasonable to avoid having the base64-encoded tokens as the hardcoded value here, and instead have something like a util method that takes the decoded base64 value (e.g. `{"sub":"w5P2WQMBQ06pyXIwSlB//A==","sid":"P8Zj1D5ITe26TteK+YuDYw==","jti":"pFm0oVCJT+CGVdXf2bK3/Q==","org":"FBS2q8/lT/2sAFKqgOiQmw==","svc":"service"}` here) and build the token out of it? That would make it slightly easier to review and understand what to expect in the future 🤔

aldexis (approved):

auth-tokens/src/test/java/com/palantir/tokens/auth/UnverifiedJsonWebTokenTests.java
pkoenig10: I think I prefer this because it ensures we're being explicit about the inputs in the tests. Having a utility method that performs the serialization is just one more place for us to introduce a bug that undermines the validity of these tests.

pkoenig10: Although I don't feel too strongly, especially if this utility is just doing Base64 encoding and not JSON serialization.

I can explore this in a FLUP.

aldexis: Yeah, I'm also wary of adding extra logic in tests to build your inputs, for the reasons you outlined. But a simple base64 encoding (and no JSON serialization as you mentioned) might be reasonable?

PR #1426, opened 2026-03-11
Make BearerToken equality comparison more efficient

lhhong:
## Before this PR

The existing `ImmutableBearerToken` generated by immutables is inefficient in a few aspects.
- On initialization, it makes a wasteful clone of the internal byte array: `this.tokenAsBytes = super.getTokenAsBytes().clone();`
- On every equality comparison, it clones the "Derived" `getTokenAsBytes` byte array again for safety: `byte[] getTokenAsBytes() { return tokenAsBytes.clone(); }`

## After this PR

==COMMIT_MSG==
Implement ImmutableBearerToken manually for performance on init / equals and remove package-private getTokenAsBytes method from BearerToken
==COMMIT_MSG==

## Possible downsides?

There is a package-private API break by removing the `getTokenAsBytes` method - but really, this shouldn't be exposed in the first place and is only there to play with immutables.


lhhong:
ABI breaks changing from abstract class to interface? Why is this a break?

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
```diff
@@ -0,0 +1,126 @@
...
+        allowedCharacters.set('-');
+        allowedCharacters.set('.');
+        allowedCharacters.set('_');
+        allowedCharacters.set('~');
+        allowedCharacters.set('+');
+        allowedCharacters.set('/');
+    }
+
+    private final String token;
+    private final transient byte[] tokenAsBytes;
```
pkoenig10: We don't claim to support Java serialization

```suggestion
    private final byte[] tokenAsBytes;
```

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
```diff
@@ -0,0 +1,126 @@
...
+    // the set of allowable characters are single bytes.
+    private static byte[] getTokenAsBytes(String token) {
+        byte[] result = new byte[token.length()];
+        for (int i = 0; i < result.length; i++) {
+            result[i] = (byte) token.charAt(i);
+        }
+        return result;
+    }
+
+    public static ImmutableBearerToken valueOf(String token) {
```
pkoenig10:
```suggestion
    static ImmutableBearerToken valueOf(String token) {
```

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
```diff
@@ -0,0 +1,126 @@
...
+        Preconditions.checkArgument(token != null, "BearerToken cannot be null");
+        Preconditions.checkArgument(!token.isEmpty(), "BearerToken cannot be empty");
+        if (!isValidBearerToken(token)) {
+            throw new SafeIllegalArgumentException(
+                    "BearerToken must match pattern", SafeArg.of("validationPattern", VALIDATION_PATTERN_STRING));
+        }
```
pkoenig10: Move the `Preconditions.checkArgument` into `isValidBearerToken`, move the validation into constructor, and remove `valueOf`.

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
```diff
@@ -0,0 +1,126 @@
...
+    // We use a hand-written getBytes() implementation for performance reasons.
+    // Note that we don't need to worry about the character set (e.g., UTF-8) because
+    // the set of allowable characters are single bytes.
+    private static byte[] getTokenAsBytes(String token) {
+        byte[] result = new byte[token.length()];
+        for (int i = 0; i < result.length; i++) {
+            result[i] = (byte) token.charAt(i);
+        }
+        return result;
+    }
```
pkoenig10: I would replace this with

```
token.getBytes(StandardCharsets.US_ASCII)
```

This will likely be faster when the string is encoded internally using `LATIN1` (which is likely the case for valid bearer tokens), since we can copy the backing byte array in a single copy instead of doing a bunch of individual reads and writes.

Would be good to capture the benchmarks before and after this change to confirm.

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
pkoenig10:
```suggestion
    private final transient byte[] tokenBytes;
```

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
```diff
@@ -0,0 +1,126 @@
...
+            }
+        }
+
+        return true;
+    }
+
+    @Override
+    @DoNotLog
+    public String toString() {
+        return getToken();
```
pkoenig10: Use fields directly internally

```suggestion
        return token;
```

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
```diff
@@ -0,0 +1,126 @@
...
+        } else if (other instanceof BearerToken bearerToken) {
+            return MessageDigest.isEqual(getTokenAsBytes(bearerToken.getToken()), tokenAsBytes);
+        }
+        return false;
+    }
+
+    @Override
+    @DoNotLog
+    public int hashCode() {
+        return getToken().hashCode();
```
pkoenig10: Use fields directly internally

```suggestion
        return token.hashCode();
```

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
```diff
@@ -0,0 +1,126 @@
...
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import java.security.MessageDigest;
+import java.util.BitSet;
+
+@DoNotLog
+final class ImmutableBearerToken extends BearerToken {
+
+    private static final String VALIDATION_PATTERN_STRING = "^[A-Za-z0-9\\-\\._~\\+/]+=*$";
+    private static final BitSet allowedCharacters = new BitSet();
```
pkoenig10: nit: I'd pull the construction of this out to a static helper method


```suggestion
    private static final BitSet ALLOWED_CHARACTERS = createAllowedCharacters();
```

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
```diff
@@ -0,0 +1,126 @@
...
+import com.palantir.logsafe.Preconditions;
+import com.palantir.logsafe.SafeArg;
+import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
+import java.security.MessageDigest;
+import java.util.BitSet;
+
+@DoNotLog
+final class ImmutableBearerToken extends BearerToken {
+
+    private static final String VALIDATION_PATTERN_STRING = "^[A-Za-z0-9\\-\\._~\\+/]+=*$";
```
pkoenig10: Let's remove this. It's only used in the error args and there's no enforcement that the pattern here matches the actual validation logic.

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
```diff
@@ -0,0 +1,126 @@
...
+        }
+        return result;
+    }
+
+    public static ImmutableBearerToken valueOf(String token) {
+        Preconditions.checkArgument(token != null, "BearerToken cannot be null");
+        Preconditions.checkArgument(!token.isEmpty(), "BearerToken cannot be empty");
+        if (!isValidBearerToken(token)) {
+            throw new SafeIllegalArgumentException(
+                    "BearerToken must match pattern", SafeArg.of("validationPattern", VALIDATION_PATTERN_STRING));
```
pkoenig10:
```suggestion
                    "Invalid bearer token"));
```

pkoenig10 (commented):
Thanks for this! Just a couple comments/suggestions.

pkoenig10:
> ABI breaks changing from abstract class to interface? Why is this a break?

https://revapi.org/revapi-java/0.28.4/differences.html#java.class.kindChanged

Class and interfaces are represented differently in the Java bytecode and affect things like method resolution, etc.

For more details, see the [Binary Compatibility chapter in the Java Language Specification](https://docs.oracle.com/javase/specs/jls/se21/html/jls-13.html).

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -19,106 +19,16 @@
...
-    @Override
-    public final boolean equals(Object other) {
-        return other instanceof BearerToken bearerToken
-                && MessageDigest.isEqual(bearerToken.getTokenAsBytes(), getTokenAsBytes());
-    }
-
-    @Override
-    @DoNotLog
-    public final int hashCode() {
-        return getToken().hashCode();
```
pkoenig10, line 122: This has the potential to be a behavior break if there are other implementations of `BearerToken` that do not override `toString, `equals`, or `hashCode`.

That being said, we generally do not support other implementations of `BearerToken`. The only internal implementation I see is that one that I presume motivated this PR.

Given this, I propose that we simply make `BearerToken` a final class and make it the singular implementation. This is a relatively simple value type - there is no good reason to allow for other implementations.

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
lhhong: Oh yes it is
manual implementation: (min, avg, max) = (31.315, 33.154, 34.582), stdev = 1.347
builtin getBytes:              (min, avg, max) = (21.076, 22.465, 25.077), stdev = 1.558

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
lhhong: Sounds good, I've seen another implementation which is so that they can mock the value - shouldn't exist as well.

auth-tokens/src/main/java/com/palantir/tokens/auth/ImmutableBearerToken.java
lhhong: Also made this `checkValidBearerToken` instead of mixing throwing illegal args and returning boolean.

auth-tokens/src/test/java/com/palantir/tokens/auth/BearerTokenTests.java
```diff
@@ -56,8 +55,7 @@ void testFromString_invalidTokens() {
         for (String invalidToken : invalidTokens) {
             Assertions.assertThatLoggableExceptionThrownBy(() -> BearerToken.valueOf(invalidToken))
                     .isInstanceOf(IllegalArgumentException.class)
-                    .hasLogMessage("BearerToken must match pattern")
-                    .hasExactlyArgs(SafeArg.of("validationPattern", "^[A-Za-z0-9\\-\\._~\\+/]+=*$"));
+                    .hasLogMessage("Invalid BearerToken");
```
pkoenig10:
```suggestion
                    .hasLogMessage("Invalid BearerToken")
                    .hasNoArgs();
```

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -45,80 +42,76 @@ public abstract class BearerToken {
...
-    private static boolean isValidBearerToken(String token) {
+    @JsonValue
+    @DoNotLog
+    public String getToken() {
+        return token;
+    }
+
+    static byte[] getTokenAsBytes(String token) {
+        return token.getBytes(StandardCharsets.US_ASCII);
+    }
```
pkoenig10: This can be private.

Let's also lazily initialize this to avoid an allocation on construction. A significant fraction, if not the majority, of bearer tokens are never used in comparisons, so we can avoid this allocation altogether in those cases.

https://github.com/palantir/auth-tokens/commit/8880340ab5133425ef7c8a6cc9be821aa8a6dcab

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -45,80 +42,76 @@ public abstract class BearerToken {
...
+
+    // Optimized validity check instead of using regular expressions
+    private static void checkValidBearerToken(String token) {
+        Preconditions.checkArgument(token != null, "BearerToken cannot be null");
+        Preconditions.checkArgument(!token.isEmpty(), "BearerToken cannot be empty");
```
pkoenig10, line 79:
```suggestion
        Preconditions.checkArgument(!token.isEmpty(), "BearerToken cannot be empty");

```

pkoenig10 (commented):
Looks good, just a couple small comments.

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -45,80 +42,76 @@ public abstract class BearerToken {
...
+        return token;
     }
 
     @Override
-    public final boolean equals(Object other) {
-        return other instanceof BearerToken bearerToken
-                && MessageDigest.isEqual(bearerToken.getTokenAsBytes(), getTokenAsBytes());
+    public boolean equals(Object other) {
+        if (other instanceof BearerToken bearerToken) {
+            return MessageDigest.isEqual(bearerToken.tokenBytes, tokenBytes);
```
pkoenig10: nit: Put this token first


```suggestion
            return MessageDigest.isEqual(tokenBytes, bearerToken.tokenBytes);
```

auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
```diff
@@ -45,80 +43,80 @@ public abstract class BearerToken {
         allowedCharacters.set('~');
         allowedCharacters.set('+');
         allowedCharacters.set('/');
+        return allowedCharacters;
+    }
+
+    private final String token;
+
+    private byte @Nullable [] tokenBytes;
```
pkoenig10: This should be `volatile`




```suggestion
    private volatile byte @Nullable [] tokenBytes;
```

pkoenig10 (dismissed):

pkoenig10:
Changing this to a break and forcing a new major version since this is an ABI break.

pkoenig10 (approved):

PR #1432, opened 2026-03-19
Remove tokenBytes

pkoenig10:
This PR copies the implementation of `MessageDigest.isEqual` and adapts it to operation on `String` values using `charAt` instead if `byte[]` values. This allows us to compare `BearerToken` values without needing to allocate and retain a duplicative `byte[]` value.

#### Before

```
Benchmark                                       Mode  Cnt   Score   Error  Units
BearerTokenBenchmark.equals_firstCharDifferent  avgt    5  65.389 ± 0.853  ns/op
BearerTokenBenchmark.equals_lastCharDifferent   avgt    5  65.404 ± 1.367  ns/op
```

#### After

```
Benchmark                                       Mode  Cnt   Score   Error  Units
BearerTokenBenchmark.equals_firstCharDifferent  avgt    5  65.001 ± 0.540  ns/op
BearerTokenBenchmark.equals_lastCharDifferent   avgt    5  65.018 ± 1.407  ns/op
```

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
@@ -21,10 +21,8 @@ import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.logsafe.DoNotLog;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
-import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.util.BitSet;
-import org.jspecify.annotations.Nullable;
 
 /** Value class representing an authentication bearer token. */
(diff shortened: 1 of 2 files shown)


aldexis (approved):

pkoenig10:
If we're willing to potentially allow the token length to be leaked when using timing attacks, then we could simplify this a bit and improve the performance.

https://github.com/palantir/auth-tokens/commit/d4df4e234719d24c0bea2f681926a80fe6d244a6

Benchmarks for this implementation are:

```
Benchmark                                       Mode  Cnt   Score   Error  Units
BearerTokenBenchmark.equals_firstCharDifferent  avgt    5  35.920 ± 1.588  ns/op
BearerTokenBenchmark.equals_lastCharDifferent   avgt    5  36.524 ± 0.696  ns/op
```

EDIT: Implemented in https://github.com/palantir/auth-tokens/pull/1433

PR #1433, opened 2026-03-19
Improve performance of token comparison

pkoenig10:
## Before this PR

`BearerToken` comparison is implemented in constant time, regardless of the length of the input value.

This imposes a slight performance penalty because we need to perform some additional work to calculate the index into the input value during the comparison.

```
Benchmark                                       Mode  Cnt   Score   Error  Units
BearerTokenBenchmark.equals_firstCharDifferent  avgt    5  65.001 ± 0.540  ns/op
BearerTokenBenchmark.equals_lastCharDifferent   avgt    5  65.018 ± 1.407  ns/op
```

## After this PR

If we are willing to sacrifice constant time comparisons when the length of the input value is different than the length of our value, then we can eliminate this performance overhead.

This feels like an acceptable tradeoff for bearer tokens. In practice, our bearer tokens are relatively large (always over 50 characters, often more like 200) and going to be one of a few different lengths. Knowing the length doesn't make it meaningful easier to forge a valid token.

```
Benchmark                                       Mode  Cnt   Score   Error  Units
BearerTokenBenchmark.equals_firstCharDifferent  avgt    5  35.727 ± 0.411  ns/op
BearerTokenBenchmark.equals_lastCharDifferent   avgt    5  35.714 ± 0.358  ns/op
```

### DIFF
--- a/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
+++ b/auth-tokens/src/main/java/com/palantir/tokens/auth/BearerToken.java
@@ -21,7 +21,6 @@ import com.fasterxml.jackson.annotation.JsonValue;
 import com.palantir.logsafe.DoNotLog;
 import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
-import java.security.MessageDigest;
 import java.util.BitSet;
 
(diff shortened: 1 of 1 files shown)


aldexis (approved):

pkoenig10:
Discussed internally in APPSEC-16281.

